PR #27, opened 2022-11-18
Upload build badge to GCS when build fails

### LINKED ISSUE #26: Failed build badge does not get uploaded anyway
If the build fails, we are able to mint the badge, but it never gets placed into the GCS bucket

jdolitsky:
Resolves #26


tstromberg (commented):
LGTM

PR #83, opened 2022-12-02
Updated PHP image to use Wolfi, added dev- image with Composer

erikaheidi:
This PR updates the current PHP image to be based on Wolfi, now that we have the php package available in the repository.

It also creates a `dev` image that has Composer and busybox installed on it. This image is useful to build applications in a multi-stage Dockerfile, with the final image being based on the distroless php image.

I do not fully understand how tagging works here, I based my changes on a few different yamls that had the `prefix` parameter set. Let me know if this looks right!


jdolitsky:
resolves #96

jdolitsky (approved):

PR #109, opened 2022-12-05
Updates in CI jobs and dependabot config

cpanato:
- use git hashes and update to latest releases
- configure dependabot

### DIFF
--- a/.github/actions/generate-matrix/action.yml
+++ b/.github/actions/generate-matrix/action.yml
@@ -10,7 +10,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
       with:
         repository: ${{ inputs.repository }}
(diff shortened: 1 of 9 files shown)


.github/actions/generate-matrix/action.yml
```diff
@@ -10,7 +10,7 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
```
cpanato, line 13: question: do we need to checkout again? because the job that calls this already do a checkout.
similar to the other reusable actions

amouat (approved):

.github/actions/generate-matrix/action.yml
amouat: Not sure. Might be worth checking in the channel.

jdolitsky: it is related to using this action in other repos. so yes we need

jdolitsky (approved):

PR #110, opened 2022-12-05
Fix READMEs for images.

amouat:
Still need more work, but this is a start.

Signed-off-by: Adrian Mouat <adrian@chainguard.dev>


README.md
```diff
@@ -1,5 +1,6 @@
 # Chainguard Images
 
+
```
cpanato, line 3: you can drop this

ran locally and and there is not difference, so we can ignore this check for now

cpanato: see slack: https://chainguard-dev.slack.com/archives/C0322APD4LU/p1670235759571799

amouat: OK, I'll try to roll back...

cpanato (approved):

PR #111, opened 2022-12-05
dont use bucket name as secret

cpanato:
there is the job `https://github.com/chainguard-images/images/blob/main/.github/workflows/presubmit-readme.yaml` that runs on PRs and inject the secret, but that will not run on Forked PRs caused the job to fail because the secret will not be injected.

we have the bucket name public and there is not needed to have that in a secret.
We can delete the secret after this PR is merged

Slack 🧵 : https://chainguard-dev.slack.com/archives/C0322APD4LU/p1670235759571799


cpanato:
will merge and cleanup the secret that will not be used anymore

PR #131, opened 2022-12-09
sdk: install gsutil and gcloud

imjasonh:
Fixes https://github.com/chainguard-images/meta/issues/22

Open to other ideas about how to do this. I wasn't sure if we wanted a Wolfi package for gsutil/gcloud since they're not OSS, or if we wanted a separate apk for them that the sdk image uses, vs putting them directly into the sdk package like I'm doing here. Feedback very welcome.

Keeping these up to date may also become a pain since they do ~weekly-ish releases, but at least since they're only used by the sdk image I think the downside of an out-of-date package is less bad than if it were in a shared place like the Wolfi repo. And anyway we only really need basic `gsutil` functionality, which doesn't change very often.

Signed-off-by: Jason Hall <jason@chainguard.dev>

### DIFF
--- a/images/sdk/configs/latest.melange.yaml
+++ b/images/sdk/configs/latest.melange.yaml
@@ -23,6 +23,7 @@ package:
       - bash
       - bubblewrap
       - tree
+      - python3
 
 environment:
   contents:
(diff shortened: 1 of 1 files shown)


jdolitsky (changes requested):
can you also add this to the wolfi melange file?

imjasonh:
> can you also add this to the wolfi melange file?

(I'm not even sure I still need this change anymore, but to answer your question) we didn't want to include these packages in Wolfi, since they're not free software.

I'm going to close this for now but we can reopen it if we decide we want these in the sdk image.

PR #139, opened 2022-12-20
Add support for multiple melange configs per image

### LINKED ISSUE #137: Support for multiple melange configs per single image
Related to https://github.com/chainguard-images/images/issues/118

jdolitsky:
Resolves #137

### DIFF
--- a/images/apko/image.yaml
+++ b/images/apko/image.yaml
@@ -1,5 +1,6 @@
 versions:
   - melange:
-      config: configs/latest.melange.yaml
+      configs:
+        - configs/latest.melange.yaml
     apko:
       config: configs/latest.apko.yaml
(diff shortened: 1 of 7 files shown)


images/apko/image.yaml
```diff
@@ -1,5 +1,6 @@
 versions:
   - melange:
-      config: configs/latest.melange.yaml
+      configs:
+        - configs/latest.melange.yaml
```
luhring, line 4: Just a quick question — if we're deciding to support both `config` and `configs` at the same time, why are we changing these instances where there's just a single config to use the multi-config option?

jdolitsky: the `multi-config` option supports a comma-separated list. In the case of a single melange file, this is still a valid list (with just a single item)

We could leave this as `config: configs/latest.melange.yaml`, but I thought it'd be easier for people copy-pasting later who need multiple melange configs to be consistent everywhere

jdolitsky: nevermind actually.. I just dropped support for the single config altogether since it isnt being used anywhere

luhring: Nice!

luhring (approved):

PR #150, opened 2022-12-25
Add netcat openbsd image

tuananh:
currently, the scaffolding chart uses trillian which rely on toolbelt/nc image. that image is amd64 only which then make the scaffolding chart requires amd64 node.

https://github.com/sigstore/helm-charts/blob/a80c33c8ca6f3c37d9b1dcb4dab14ca625f77831/charts/trillian/values.yaml#L160

### DIFF
--- a/images/netcat-openbsd/configs/latest.apko.yaml
+++ b/images/netcat-openbsd/configs/latest.apko.yaml
@@ -1,0 +1,21 @@
+contents:
+  repositories:
+    - https://dl-cdn.alpinelinux.org/alpine/edge/main
+    - https://dl-cdn.alpinelinux.org/alpine/edge/community
+  packages:
+    - ca-certificates-bundle
+    - busybox
+    - netcat-openbsd
+accounts:
+  groups:
+    - groupname: nonroot
+      gid: 65532
+  users:
+    - username: nonroot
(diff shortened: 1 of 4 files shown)


images/netcat-openbsd/configs/latest.apko.yaml
```diff
@@ -0,0 +1,26 @@
+contents:
+  repositories:
+    - https://dl-cdn.alpinelinux.org/alpine/edge/main
+    - https://dl-cdn.alpinelinux.org/alpine/edge/community
+  packages:
+    - ca-certificates-bundle
+    - busybox
+    - netcat-openbsd
```
jdolitsky, line 8: i think we are trying to prevent new alpine-based images in this repo.

Maybe we can look into packaging netcat in wolfi

jdolitsky (changes requested):

images/netcat-openbsd/configs/latest.apko.yaml
tuananh: lemme try adding netcat-openbsd PR at wolfi repo

jdolitsky:
closing as old/inactive. please re-open if necessary

PR #152, opened 2022-12-26
Add a go 1.20 prerelease image.

dlorenc:
I'm not 100% sure I did the tags part right here.


images/go/configs/latest-1.20.apko.yaml
```diff
@@ -0,0 +1,19 @@
...
+  packages:
+    - ca-certificates-bundle
+    - busybox
+    - build-base
+    - git
+    - openssh
+    - go-1.20
+
+environment:
+  GODEBUG: tarinsecurepath=0,zipinsecurepath=0
```
dlorenc, line 15: Also I took the liberty of adding these environment variables, they enable some new security features in go 1.20!

images/go/configs/latest-1.20.apko.yaml
```diff
@@ -0,0 +1,19 @@
+contents:
```
imjasonh, line 1: Maybe just 1.20.apko.yaml? We don't need the latest part I think.

dlorenc: Fixed!

jdolitsky (approved):

PR #154, opened 2022-12-27
Add a prometheus image.

dlorenc:


amouat (approved):
Are the ports specified in the prometheus.yml? Might be worth adding a link to the example config and specifying what the default ports are here.

dlorenc:
> Are the ports specified in the prometheus.yml? Might be worth adding a link to the example config and specifying what the default ports are here.

PR #155, opened 2022-12-27
Add a curl image.

dlorenc:


imjasonh (approved):

images/curl/configs/latest.apko.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+accounts:
+  groups:
+    - groupname: curl
+      gid: 65532
+  users:
+    - username: curl
+      uid: 65532
+      gid: 65532
+  run-as: 65532
```
imjasonh, line 20: Needs archs here or else it defaults to `all`. I'm not sure why CI didn't catch this though 🤔

PR #156, opened 2022-12-27
Add etcd image.

dlorenc:


images/etcd/configs/latest.apko.yaml
```diff
@@ -0,0 +1,16 @@
+contents:
+  repositories:
+    - https://packages.wolfi.dev/os
+  keyring:
+    - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+  packages:
+    - etcd
+    - ca-certificates-bundle
+    - glibc
```
amouat, line 9: We shouldn't need to specify glibc.

amouat (changes requested):
Is this something we should worry about? 

https://github.com/etcd-io/etcd/blob/main/Dockerfile-release.amd64#L13

I see the reference is to "alpine", but the FROM is for a distroless image.

Also, there is a _lot_ more in the bitnami image regarding start-up, but most of it seems concerned with mapping env vars to config.

images/etcd/configs/latest.apko.yaml
dlorenc: Double checked and we do need to keep glibc here - the builds are dynamic on purpose.

dlorenc:
> Is this something we should worry about?
> 
> https://github.com/etcd-io/etcd/blob/main/Dockerfile-release.amd64#L13
> 
> I see the reference is to "alpine", but the FROM is for a distroless image.
> 
> Also, there is a _lot_ more in the bitnami image regarding start-up, but most of it seems concerned with mapping env vars to config.

I think this comment is out of date and they probably meant distroless. But distroless does include an nsswitch.conf now, as does Wolfi.

jdolitsky (approved):
cannot speak to Adrian's comments, but the repo changes lgtm

amouat:
Yeah, it just needs the glibc entry removed now that it's in wolfi. 

Should we mark this experimental though? etcd is a critical component so we might want to be careful.

jdolitsky:
how to do so if we wanna: https://github.com/chainguard-images/images/blob/main/images/bazel/image.yaml#L1

amouat (approved):

amouat:
LGTM, but not sure if you're still testing @dlorenc so I'll let you merge.

PR #157, opened 2022-12-27
Add a redis image.

dlorenc:


amouat (changes requested):
This is a point for all the database containers, so I apologise for picking on redis, but I think we should consider adding an entrypoint script to "fix" permission issues e.g: https://github.com/docker-library/redis/blob/master/7.0/docker-entrypoint.sh#LL12C2-L12

This does suggest we support "entrypoint", so we may want to discuss further.

It would also bloat the image with a shell.

Also note this image is running as root. I think we need to quickly develop a "checklist" for our images.

jdolitsky (approved):
cannot speak to Adrian's comments, but the repo changes lgtm

dlorenc:
I don't love the entrypoint "magic" - it feels like a tradeoff between security and usability (bloating the container and introducing unpredictable behavior in order to make it easier for folks to get running as a random user).

But agree completely this shouldn't run as root. Let me try to take another pass with a predictable user ID and directory.

amouat:
Yeah, I was thinking about this overnight. Let's forget about the entrypoint, but change it from root and add some documentation on how to set up a volume for Redis with the correct owner.

images/redis/README.md
```diff
@@ -0,0 +1,36 @@
...
+docker run -p 6379:6370--rm cgr.dev/chainguard/redis
+1:C 27 Dec 2022 16:42:20.647 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
+1:C 27 Dec 2022 16:42:20.647 # Redis version=7.0.7, bits=64, commit=00000000, modified=0, pid=1, just started
+1:C 27 Dec 2022 16:42:20.647 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
+1:M 27 Dec 2022 16:42:20.648 * monotonic clock: POSIX clock_gettime
+1:M 27 Dec 2022 16:42:20.648 * Running mode=standalone, port=6379.
+1:M 27 Dec 2022 16:42:20.648 # Server initialized
+1:M 27 Dec 2022 16:42:20.650 * Ready to accept connections
+```
+
```
amouat, line 28: Can we add a second example with a volume for the datadir, where we first run a chmod or chown on the host directory?

amouat (approved):
LGTM. If we add a second example I think we've covered everything.

images/redis/README.md
dlorenc: Sorry, do you have an example of what you mean? The volume directive in apko does a good job here so even if you mount that `/data` directory to a hostpath it "just works".

I added an example in here!

amouat: Yeah, if I do something like:

```
$ mkdir mydata
$ docker run -d -v $PWD/mydata:/data cgr.dev/chainguard/redis
```

Does it work? Or do I need to do a chmod/chown on `mydata`? Not sure if you need to add data and force a save to test this.

dlorenc: Yea that one "just works"! Covered in an example.

PR #163, opened 2022-12-29
Change wolfi-base to use cmd.

amouat:
Entrypoint causes confusion and is uneeded as commands are interpreted with sh -c by default.


images/wolfi-base/configs/latest.apko.yaml
```diff
@@ -7,8 +7,7 @@ contents:
     - ca-certificates-bundle
     - wolfi-base
 
-entrypoint:
-  command: /bin/sh -l
+cmd: /bin/sh -l
```
imjasonh, line 10: Is there an accompanying change we can make to a test.sh to document this change, and prevent regressions? This is a subtle change and I don't want to accidentally undo it in 3 months.

amouat: @dlorenc what did it break for you?

We can add tests that:
 - running without any command works without error (-it should result in shell, but that's harder to test)
 - running with a shell command e.g. `ls` works
 
But both of those should work with the entrypoint as well.

amouat: I'm still not sure if we want the `-l` btw.

amouat: Actually these tests might be enough. With the current image:

```
docker run cgr.dev/chainguard/wolfi-base:latest ls
/bin/sh: can't open 'ls': No such file or directory
```

😳 

How do tests work in the monorepo? Do I have to edit something to pick them up or will it automatically look for test.sh?

jdolitsky: if test.sh executable is present, it will try to run it

jdolitsky (approved):

amouat:
Going to merge as I think the current version is broken and we have tests which I think address @imjasonh's comment.

images/wolfi-base/configs/latest.apko.yaml
johnfosborneiii: > Is there an accompanying change we can make to a test.sh to document this change, and prevent regressions? This is a subtle change and I don't want to accidentally undo it in 3 months.

You could keep entrypoint and append `exec "$@"'` which would keep backwards compat and then run the commands passed by the user. It could also allow for some pre-processing if needed in the future.

amouat: I don't think there's any backwards compat to keep tbh. I'm not sure that would work anyway due shell stuff.

PR #164, opened 2022-12-29
Add a deno image.

dlorenc:
This is marked as experimental. Deno doesn't actually require a base image to run, so we should also show some multi-stage build examples here where you use deno compile then a static base.


images/deno/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+The image is available on `cgr.dev`:
+
+```
+docker pull cgr.dev/chainguard/deno:latest
+```
+
+## Usage Example
+
+Navigate to the [`example/`](./example/) directory:
```
amouat, line 17: I like this example, but the problem is we have a monorepo now, and I don't want people to have to check out everything :/

I'm not sure what to suggest. The Dockerfile is small enough to inline here and I guess it could download the ts?

amouat (approved):
I'd like to make the example more accessible, but the image looks good!

dlorenc:
Ah, looks like `deno compile` doesn't generate static binaries, they require a few things:

```shell
$  ldd /app/webserver
	linux-vdso.so.1 => linux-vdso.so.1 (0x0000ffffa6480000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x0000ffffa2780000)
	libm.so.6 => /lib/libm.so.6 (0x0000ffffa26d0000)
	libc.so.6 => /lib/libc.so.6 (0x0000ffffa2530000)
	/lib/ld-linux-aarch64.so.1 => /lib/ld-linux-aarch64.so.1 (0x0000ffffa6443000)
```

Our glibc images don't have `libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x0000ffffa2780000)`. We do have that packaged, but it's only available in our gcc images which are huge. If there's interest we could add a dynamic image for deno apps.

images/deno/README.md
dlorenc: Done!

amouat:
>Our glibc images don't have libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x0000ffffa2780000). We do have that packaged, but it's only available in our gcc images which are huge. If there's interest we could add a dynamic image for deno apps.

Makes sense to me. Is this a common set of libraries? I wonder if there's a name for it or something...

PR #165, opened 2022-12-30
Add 'cc' image.

dlorenc:
This is a clone of the distroless/cc image, which we don't really have today. Rust, deno, and a few other binary types need this library.


amouat:
Oh. This is interesting. So it has more libs than glibc-dynamic? I'm wondering how we tell people which images to use :/

dlorenc:
> Oh. This is interesting. So it has more libs than glibc-dynamic? I'm wondering how we tell people which images to use :/

Yeah - distroless has this setup too - there's "static", "base", and "cc".

I just tried to get a rust app running on glibc-dynamic and it fails because it's missing `libgcc_s.so.1`. Deno ran into a similar issue yesterday. The only image we have that contains this today is gcc, but that's massive because it contains the entire compiler.

amouat:
Ok. Sounds good to me then. We could do with an article on this.

amouat (approved):

PR #166, opened 2022-12-30
Flip the default go image from alpine/musl to wolfi/glibc!

dlorenc:
The old one is still available but at cgr.dev/chainguard/go:musl-latest.

This was generated by copying the old latest-glibc.apko.yaml -> latest.apko.yaml and moving the old latest.apko.yaml -> latest-musl.apko.yaml (the diff looks a little weird).


luhring (approved):
This is a breaking change, but it also seems inevitable

images/go/configs/latest-musl.apko.yaml
```diff
@@ -19,13 +20,8 @@ accounts:
...
       gid: 65532
+  # TODO: this should run as 65532
   run-as: 0
-
 entrypoint:
   command: /usr/bin/go
-
 cmd: help
-
-archs:
```
imjasonh, line 29: We're going to lose a lot of archs if we do this. Not sure if folks are using them though, without some more querying.

dlorenc: Ack. I'm personally fine with it for these images. Happy to hold if we want to wait longer though.

kimsterv:
> This is a breaking change, but it also seems inevitable

Who are we breaking and how should we be communicating these changes? cc @amouat

amouat:
> > This is a breaking change, but it also seems inevitable
> 
> Who are we breaking and how should we be communicating these changes? cc @amouat

I'm not sure. I think it's pretty limited in breaking people. I'm assuming it compiles against the system libs and therefore you get a glibc binary instead of the previous musl one, which may not work in certain environments.

It will break people using archs other than arm64 and amd64, but I suspect there's so few of those we can ignore it.

I would make the change but add something to the page explaining the change and how to get the musl image.

dlorenc:
> > > This is a breaking change, but it also seems inevitable
> > 
> > 
> > Who are we breaking and how should we be communicating these changes? cc @amouat
> 
> I'm not sure. I think it's pretty limited in breaking people. I'm assuming it compiles against the system libs and therefore you get a glibc binary instead of the previous musl one, which may not work in certain environments.
> 
> It will break people using archs other than arm64 and amd64, but I suspect there's so few of those we can ignore it.
> 
> I would make the change but add something to the page explaining what happened and how to get the musl image.

Added a sentence to the page.

PR #183, opened 2023-01-03
ruby: use irb as entrypoint

imjasonh:
This matches the `ruby` image on Dockerhub:

```
$ docker run --rm -it ruby
irb(main):001:0>
```

and now, with this change:

```
$ docker run --rm -it image.tar
irb(main):001:0>
```

Signed-off-by: Jason Hall <jason@chainguard.dev>

### DIFF
--- a/images/ruby/configs/latest-3.0.apko.yaml
+++ b/images/ruby/configs/latest-3.0.apko.yaml
@@ -26,8 +26,7 @@ paths:
 work-dir: /work
 
 entrypoint:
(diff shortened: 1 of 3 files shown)


dlorenc (approved):

luhring:
Would it be better to make it a command instead of entrypoint?

amouat:
Yeah @luhring is right. We should just set `cmd` and not `entrypoint`.

PR #190, opened 2023-01-04
Move ruby to use cmd not entrypoint and add test.

amouat:


tuananh:
i like this change. it's least surprise when switching from the official ruby image.

luhring:
@amouat Just double-checking about the test mentioned in the PR title

amouat:
good point @luhring ! One sec...

amouat:
The test does very little - just runs with irb and ruby with `--version`. It will pick up if the entrypoint is changed (unless something clever is done with a script).

luhring (approved):

jdolitsky (approved):

PR #192, opened 2023-01-04
Remove melange builds for apko and melange images

jdolitsky:
The "from HEAD" images are now located at https://github.com/wolfi-dev/tools:
```
ghcr.io/wolfi-dev/apko:latest
ghcr.io/wolfi-dev/melange:latest
```


images/melange/image.yaml
```diff
@@ -1,4 +1,8 @@
 versions:
+  - apko:
+      config: configs/0.apko.yaml
```
imjasonh, line 3: Will this also tag them with :0?

jdolitsky: yes. and actually, even if we rename this, the way tagging works with extractTagsFrom and the version stemming, 0 would be added as a version tag anyway, since it is the major version technically

jdolitsky: we would have to add an option to ignore major / single digit tags inside apko

imjasonh (approved):

PR #194, opened 2023-01-05
monopod to enforce README per image, regen all README.md files

jdolitsky:
This allows us to enforce READMEs per each image to show all available tags etc.


jdolitsky:
outstanding TODO to simplify some of the duplicated code here

amouat:
LGTM, but it might be worth getting someone who knows Go and our workflow better to take a look before merging.

rawlingsj (approved):

PR #202, opened 2023-01-07
Move graal from experimental to main.

dlorenc:
We got some feedback that this approach works and we can remove experimental!


luhring (approved):

tisonkun:
Thank you! I'm releasing [v1.2.0 for korandoru/hawkeye](https://github.com/korandoru/hawkeye/releases/tag/v1.2.0), whose Native Image Docker image is based on this image. Its usage is verified on my projects that use it.

tisonkun:
Does this PR mean that we can use a stable tag rather than `latest` to refer the image now?

dlorenc:
> Does this PR mean that we can use a stable tag rather than `latest` to refer the image now?

Normally it would, but there's not really anything in this image for us to anchor the versioning off of (for example, with Python we'd generate a stable tag for something like 3.11.1). For this image it's just glibc and a few other packages, so we don't have something to anchor the tag to.

If there's a specific tagging scheme you'd like to see here we can definitely work it in!

tisonkun:
@dlorenc That's reasonable. Then I'll keep the current way unless we find something needs a tag different than `latest`. I may expect nothing will be changed in this image except version bump of glibc, libz, etc. to make to work with latest GraalVM.

dlorenc:
Thanks, that's our intention here too!

PR #203, opened 2023-01-09
Switch the rabbitmq group to 999 and add bash which is required for t…

dlorenc:
…he entrypoint.

### DIFF
--- a/images/rabbitmq/configs/latest.apko.yaml
+++ b/images/rabbitmq/configs/latest.apko.yaml
@@ -6,33 +6,34 @@ contents:
   packages:
     - rabbitmq-server
     - wolfi-baselayout
+    - bash
 
 paths:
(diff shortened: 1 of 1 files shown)


jdolitsky (approved):

amouat:
@dlorenc just to be clear, rabbitmq expects the user to be uid 999?

jdolitsky:
needs rebase: 
```
The head commit for this pull_request event is not ahead of the base commit. Please submit an issue on this action's GitHub repo.
```

jdolitsky (approved):

dlorenc:
> @dlorenc just to be clear, rabbitmq expects the user to be uid 999?

The rabbitmq operator, which is the primary method of installing rabbitmq in clusters seems to have that expectation :(

PR #210, opened 2023-01-10
Remove Go module cache from sdk image to shrink size

jdolitsky:


kaniini (approved):

images/sdk/configs/latest.melange.yaml
```diff
@@ -58,13 +58,9 @@ pipeline:
       git clone https://github.com/chainguard-dev/apko.git
       (cd apko && make apko install)
 
+      # move all Go-built stuff into /usr/bin, remove uneeded files
+      mv "${GOPATH}/bin/*" "${DESTDIR}/usr/bin"
```
kaniini, line 62: I don't think we need this part

jdolitsky: just pushed up a new version. trying to get goimports into the package

kaniini: Why?  It's not a hard requirement for building, and I don't think people really use this image to build apko and melange itself, but rather to make use of them.

jdolitsky: > I don't think people really use this image to build apko and melange itself

i do personally :)

kaniini: Carry on, then :)

PR #227, opened 2023-01-18
add new wolfi based nginx and change the default latest tag to point …

rawlingsj:
…to it

Signed-off-by: James Rawlings <jrawlings@chainguard.dev>


rawlingsj:
I don't think I've got this PR quite right yet, not 100% sure about the tags.  Also as it's changing the default latest nginx image to be wolfi based it would be good to get some eyes on it.

jdolitsky:
Failure due to nginx not starting up properly due to permissions:
```
nginx: [alert] could not open error log file: open() "/var/lib/nginx/logs/error.log" failed (13: Permission denied)
2023/01/19 23:22:43 [emerg] 8#8: unknown log format "main" in /etc/nginx/nginx.conf:41
nginx-smoketest-1674170563
```

jdolitsky (approved):
approving in advance, considering we can fix the paths stuff

rawlingsj:
Not too sure yet whats changed as this was working and passing CI - looking into it

rawlingsj:
The CI failure seems to be apko using the wrong nginx package from wolfi, the logs say
```
Jan 19 23:42:04.062 [INFO] [arch:x86_64] installing nginx (1.23.3-r0)
```
when it should be `1.23.3-r1`

jdolitsky:
we just need to re-run CI, there was a regression in apko

PR #236, opened 2023-01-20
Make glibc/wolfi version latest for busybox and git

jdolitsky:


jdolitsky:
nvm nvm - the latests versions are needed to be multiarch for now. Can revisit once wolfi supports s390x+ppc64le

PR #239, opened 2023-01-21
Rust-ify the curl image!

dlorenc:
I think this is the right way to force curl to select the rustls backend.


dlorenc:
cc @kaniini

kaniini (approved):

PR #244, opened 2023-01-25
Revert "rabbitmq: switch uid and guid to be consistent with other cha…

rawlingsj:
…inguard images"

This reverts commit 4588f0119aabefdbc88cd1a7191aee301abc0ece.

It seems these GID and UID do need to be 999 https://sourcegraph.com/github.com/rabbitmq/cluster-operator/-/blob/internal/resource/statefulset.go?L541


luhring (approved):
This would be a cool use of the new Melange "vars" feature (only... in apko :D )

PR #250, opened 2023-01-30
go: add support for 1.18, extract tags for 1.18 and 1.19

imjasonh:
This adds a Go 1.18 image, which will be tagged with `1.18`, `1.18.9`, and `1.18.9-r3`, but explicitly _not_ `1`.

This also changes `:latest` to have it extract tags from the `go-1.19` package, which includes `1`, `1.19`, `1.19.5`, `1.19.5-r3`. Extracting tags from `go` didn't seem to work, since `go` is actually [`provide`d by `go-1.19`](https://github.com/wolfi-dev/os/blob/3bb0c5f0bab78a062b7459094476a78d0ebddc3c/go-1.19.yaml#L15) and not a real package. I'm not sure we would have wanted that anyway, since it would have tagged it as `:1.19.999` instead of the correct `go-1.19` version.

`1.20` is unaffected.


jdolitsky (approved):
👍

PR #254, opened 2023-02-03
Increase speed of policy-check by using Wolfi packages

jdolitsky:
This shaves the time of the policy-check step from 3m12s to 25s


tuananh:
may i ask why don't we just setup an action to fetch binary directly from release? https://github.com/sigstore/policy-controller/releases/tag/v0.6.2

jdolitsky:
> may i ask why don't we just setup an action to fetch binary directly from release? https://github.com/sigstore/policy-controller/releases/tag/v0.6.2

That would also work as far as speed is concerned.

We had a conversation at Chainguard this week about dogfooding Wolfi packages instead of using various "x-installer" actions. The action installer for Grype last week stopped working which prompted us to start looking into using Wolfi packages. This also has the benefit of being able to automatically receive updates of these tools, with still having ability to pin if necessary (e.g. `policy-controller-tester=0.6.2-r0`).

This is a bit of an experiment, but imagine this type of functionality baked into a general GitHub Action maintained by the Wolfi project.

.github/actions/policy-check-image/action.yml
```diff
@@ -5,20 +5,44 @@ inputs:
...
       shell: bash
       run: |
-        (cd policy-controller/ && make policy-tester && mv policy-tester /tmp/policy-tester)
-    - uses: imjasonh/setup-crane@v0.1
+        trap "rm -rf custom-image.tar custom-image.apko.yaml" EXIT
+
+        cat >./custom-image.apko.yaml <<EOL
+        contents:
+          packages:
+            - glibc # TODO: fix policy-controller-tester package to not need this or automatically require it
```
jdolitsky, line 16: TODO should be dropped when new policy-controller-tester is built: https://github.com/wolfi-dev/os/pull/530

jdolitsky: hm.. doesnt seem to fix the issue

jdolitsky:
closing for now

imjasonh (approved):

jdolitsky:
re-opening now as we are running into https://github.com/sigstore/policy-controller/issues/608

PR #267, opened 2023-02-09
allow optional override of checkout ref when generating matrix if def…

rawlingsj:
…ault github action vars dont exist, when using `pull_request_target` for example

Signed-off-by: James Rawlings <jrawlings@chainguard.dev>


rawlingsj:
the approach needs to be checked with @jdolitsky

PR #281, opened 2023-02-17
add dev for java images

strongjz:


images/jdk/configs/latest-dev.apko.yaml
```diff
@@ -0,0 +1,39 @@
...
+environment:
+  LANG: en_US.UTF-8
+  JAVA_HOME: /usr/lib/jvm/openjdk
+
+paths:
+  - path: /home/build
+    type: directory
+    uid: 65532
+    gid: 65532
+    permissions: 0o755
```
jedsalazar, line 35: is using octal format the preferred way here?

images/jdk/configs/latest-dev.apko.yaml
```diff
@@ -0,0 +1,39 @@
+contents:
+  repositories:
+    - https://packages.wolfi.dev/os
+  keyring:
+    - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+  packages:
+    - ca-certificates-bundle
```
imjasonh, line 7: Are we supposed to include `apk-tools` in `-dev` images?

images/jdk/configs/openjdk-11-dev.apko.yaml
```diff
@@ -0,0 +1,39 @@
+contents:
+  repositories:
+    - https://packages.wolfi.dev/os
+  keyring:
+    - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+  packages:
+    - ca-certificates-bundle
```
imjasonh, line 7: ditto

jedsalazar (commented):
lgtm - just one question

images/jdk/image.yaml
```diff
@@ -10,4 +10,16 @@ versions:
       config: configs/openjdk-11.apko.yaml
       extractTagsFrom:
         package: openjdk-11
+        prefix: openjdk-
+  - apko:
+      config: configs/latest-dev.apko.yaml
+      extractTagsFrom:
+        package: openjdk-17-dev
```
imjasonh, line 17: There doesn't appear to be a Wolfi package named `openjdk-17-dev`, I think we want this to be `package: openjdk-17`, and `prefix: dev-`?

That will give us `:dev-openjdk-17-11.0.18-r1`, is that right?

strongjz: fixed

images/jdk/configs/latest-dev.apko.yaml
strongjz: yes? https://github.com/chainguard-dev/apko/blob/main/examples/nginx.yaml#L32

images/jdk/image.yaml
rawlingsj: I think @imjasonh comment is still valid, there isn't a package called `openjdk-17-dev` should this be `openjdk-17`?

images/jre/image.yaml
```diff
@@ -3,13 +3,27 @@ versions:
       config: configs/latest.apko.yaml
       extractTagsFrom:
         package: openjdk-17
-        prefix: openjdk-
+        prefix: openjdk-jre
```
rawlingsj, line 6: FWIW the `-jre` suffix was dropped in another PR two weeks ago https://github.com/chainguard-images/images/pull/256/files#diff-3313beffc0b154f1127c0db56645507d074031c287e20c280555f1d9e6e87d36L6.  I think that was driven by a view that it was redundant given `jre` appears in the image name.  @amouat @patflynn is that right?

patflynn: yep that's correct.

images/jre/image.yaml
```diff
@@ -3,13 +3,27 @@ versions:
...
       tags:
         - openjdk-jre-17 # TODO: deprecate this old tag
   - apko:
       config: configs/openjdk-11.apko.yaml
       extractTagsFrom:
         package: openjdk-11
-        prefix: openjdk-
+        prefix: openjdk-jre
```
rawlingsj, line 13: Same as comment above `-jre` was dropped previously

rawlingsj (approved):

PR #283, opened 2023-02-17
add dev for go

strongjz:


images/go/image.yaml
```diff
@@ -9,3 +9,15 @@ versions:
...
+      config: configs/1.19-dev.apko.yaml
+      extractTagsFrom:
+        package: go-1.19
+        prefix: dev-
+        exclude:
+          - 1
+  - apko:
+      config: configs/latest-dev.apko.yaml
+      extractTagsFrom:
+        prefix: dev-
```
imjasonh, line 22: (not blocking)

This will produce `:dev-1.20.1-r0` which is also tagged `:latest-dev`?

We might want to standardize on whether `dev` comes first or last... should we have a `suffix:` option as well?

imjasonh (approved):

images/go/image.yaml
amouat: Last. It should come last.

PR #286, opened 2023-02-21
update java/go latest with variant version

strongjz:


amouat (approved):

images/go/image.yaml
```diff
@@ -13,11 +13,11 @@ versions:
       config: configs/1.19-dev.apko.yaml
       extractTagsFrom:
         package: go-1.19
-        prefix: dev-
         exclude:
           - 1
   - apko:
       config: configs/latest-dev.apko.yaml
       extractTagsFrom:
-        prefix: dev-
```
jdolitsky, line 22: without a prefix, this will overwrite the latest one. we can just remove `extractTagsFrom:` altogether

jdolitsky (changes requested):

images/go/image.yaml
```diff
@@ -13,11 +13,11 @@ versions:
...
         package: go-1.20
+      tags:
+        - go-1.20-dev
```
jdolitsky, line 23: why have a tag with "go-" in the name? this is the go image. Are you looking for a tag suffix basically?

jdolitsky (changes requested):

images/go/image.yaml
strongjz: good catch, ill remove the go-

images/go/image.yaml
strongjz: fixed

images/jdk/image.yaml
```diff
@@ -14,11 +14,12 @@ versions:
   - apko:
       config: configs/latest-dev.apko.yaml
       extractTagsFrom:
-        package: openjdk-11
+        package: openjdk-17
         prefix: openjdk-
```
jdolitsky, line 18: like the other comment, this is going to overwrite latest above. should just remove `extractTagsFrom`. This was already broken before your changes it seems

jdolitsky (changes requested):

images/jdk/image.yaml
strongjz: updated

jdolitsky (approved):

PR #290, opened 2023-02-22
Attempt at PR Template with checklist

amouat:
Not sure if this is correct!


.github/PULL_REQUEST_TEMPLATE/checklist.md
```diff
@@ -0,0 +1,31 @@
+# Checklist
```
patflynn, line 1: could we use tasklist style markdown so that the author has to check the boxes to confirm?  

Will this be applied to all PRs?  If so perhaps it can start with a header explaining to delete the checklist if this is not for a new image?

Issue #298, opened 2023-02-23
ERROR: Unable to lock database: Permission denied - trying to install apks in the new dev varient 


rawlingsj:
```sh
docker run -ti --entrypoint bash cgr.dev/chainguard/jre:latest-dev
bash-5.2$ cat
bash: cat: command not found
bash-5.2$ ls
bash: ls: command not found
bash-5.2$ apk add busybox
ERROR: Unable to lock database: Permission denied
ERROR: Failed to open apk database: Permission denied
```

jdolitsky:
`cat` and `ls` come from busybox, so we are missing that package

as far as the permissions error, solved by running as root:
```
$ docker run -u root -ti --entrypoint bash cgr.dev/chainguard/jre:latest-dev
bash-5.2# apk add busybox
fetch https://packages.wolfi.dev/os/aarch64/APKINDEX.tar.gz
(1/1) Installing busybox (1.36.0-r0)
Executing busybox-1.36.0-r0.trigger
OK: 179 MiB in 21 packages
```

but we should add something to dev variants to allow modifying the apk db directory

rawlingsj:
Ah didn't mean to confuse with adding the cat and ls, I was aiming to show why I was installing busybox and getting the permission error.  Running as root fixed it thanks.

PR #299, opened 2023-02-23
add busybox to some dev images

patflynn:


patflynn:
should fix the missing cp/ls commands for https://github.com/chainguard-images/images/issues/298

jdolitsky:
is this still needed as #298 is closed?

patflynn:
:man_shrugging: we are going to have busybox by default with wolfi-base in dev images, so this change would have the jre dev images work the same way as most of the others currently in the repo.

jdolitsky (approved):

PR #329, opened 2023-03-07
skip policy check while testing upgrade to latest version

rawlingsj:


rawlingsj:
Not needed - all pipelines are in fact green, we had bogus alerts saying otherwise.

PR #346, opened 2023-03-15
add busybox

strongjz:
add busybox to python dev images

### DIFF
--- a/images/python/image.yaml
+++ b/images/python/image.yaml
@@ -26,19 +26,13 @@ options:
     contents:
       packages:
         add:
+          - build-base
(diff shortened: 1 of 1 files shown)


jdolitsky:
"busybox" already brought in in globals: https://github.com/chainguard-images/images/blob/main/globals.yaml#L9

 Python packages should be updated to look like following:
```
 python-3-11-dev:
    contents:
      packages:
        add:
          - python-3.11-dev
          - build-base
          - py3.11-pip
```

jdolitsky (approved):

PR #353, opened 2023-03-17
Feature: Start to authenticate with Chainguard registry.

mattmoor:
:gift: This change adds setup-chainctl to start authenticating with our registry.

A subsequent change will start to take advantage of this, but we want to try authenticating as the first baby step.

/kind feature


jdolitsky (approved):

patflynn:
I still can't get over how easy that is.

PR #358, opened 2023-03-18
Move the nats and nsc binaries to the nats-dev image using build-opti…

dlorenc:
…ons.

The nats image in the nats repo contains these by default, but the official Docker image does not.

Fixes:

Related:

### DIFF
--- a/images/nats/image.yaml
+++ b/images/nats/image.yaml
@@ -8,3 +8,11 @@ versions:
         - suffix: -dev
           options:
             - dev
+            - nats-dev
+options:
+  nats-dev:
+    contents:
+      packages:
+        add:
+          - nats
+          - nsc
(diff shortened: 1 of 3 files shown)


images/nats/image.yaml
```diff
@@ -8,3 +8,10 @@ versions:
         - suffix: -dev
           options:
             - dev
+options:
+  dev:
```
jdolitsky, line 12: lets call this "nats-dev", then add it the the list above under dev, otherwise I think it will override the global dev which has apk-tools etc

jdolitsky (changes requested):

images/nats/image.yaml
dlorenc: Good call!

jdolitsky (approved):

PR #368, opened 2023-03-21
Add -dev variant to static image

charlieegan3:
Plan is to replace `gcr.io/distroless/static:debug` here https://github.com/open-policy-agent/opa/blob/main/Makefile#L363

Fixes:

Related: 

### Pre-review Checklist

#### For new image PRs only

- [ ] Image is marked experimental or stable as appropriate
- [ ] The last two minor versions are available
- [ ] ENTRYPOINT
- [ ] CMD:
- [ ] Add annotations e.g:
```
annotations:
```
- [ ] Include tests, at the very least a basic smoke test.


charlieegan3:
> IMPORTANT: 'image-request' tag has been applied if this PR is adding any images, including new versions or variants

I don't think that I can set the labels, so might need some help there.

images/static/image.yaml
```diff
@@ -1,5 +1,9 @@
 versions:
   - apko:
       config: configs/latest.apko.yaml
+      subvariants:
+        - suffix: -dev
+          options:
+            - dev
   - apko:
       config: configs/latest-glibc.apko.yaml
```
jdolitsky, line 9: if we could would be great to have this on the glibc version too

imjasonh:
I might be missing some context, but what is a dev variant of `static` supposed to achieve? As I understand it, `static:latest-dev` would be identical to `static:latest` with the PR in its current state. Typically dev variants have other stuff in them to make them useful to developers.

The nearest equivalent to `gcr.io/distroless/static:debug` would probably be something like `cgr.dev/chainguard/busybox`, which has no package manager, only a shell. If you want a package manager, you can use `wolfi-base`, which has `busybox` and `apk`.

charlieegan3:
Thanks for those details Jason, much appreciated. I think that we can use busybox in that case. I'll close this for now!

PR #369, opened 2023-03-21
push to cgr.dev

imjasonh:
This will need a change to `setup-chainctl` to correctly configure creds for `cgr.dev`, but once that's in we should be good to go.

cc @mattmoor

### DIFF
--- a/.github/actions/release-image/action.yml
+++ b/.github/actions/release-image/action.yml
@@ -323,7 +323,7 @@ runs:
             crane cp "${REF}" "${tag}"
             # For public images, copy them over to Chainguard registry
             if [[ "${{ inputs.chainguardNamespace }}" != "" ]]; then
-              CGR_REF="registry.enforce.dev/${{ inputs.chainguardNamespace }}/$(basename "${tag}")"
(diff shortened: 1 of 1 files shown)


dlorenc:
Will we also be able to remove the random sleeps?

jdolitsky:
I'm looking into the CI failures

.github/actions/release-image/action.yml
```diff
@@ -296,6 +296,9 @@ runs:
       if: inputs.chainguardIdentity != ''
       with:
         identity: ${{ inputs.chainguardIdentity }}
+    - run: |
```
jdolitsky, line 299: `if: inputs.chainguardIdentity != ''`

mattmoor, line 299: We want the `if` check on the identity (or drop it entirely)

mattmoor:
@dlorenc re: sleeps, we need to reconfigure the workflow to push to `cgr.dev` during the main part of the workflow (and copy to ghcr.io/chainguard-images at the end), and then we should be able to 🤞

mattmoor (approved):

jdolitsky (approved):
@imjasonh merge at will

PR #374, opened 2023-03-23
Ability to mark certains tests with tags (e.g. k8s etc)

jdolitsky:
This sets up a cron/workflow dispatch job to run certain tests in the context of a kind cluster / local registry. Tests scripts need to contain the string `# monopod:tag:functional` and they will be ignored by the regular workflows, and get picked up by the functional test workflow. Currently only applied to fluent-bit (ty @found-it for test example), and it is failing.


images/fluent-bit/tests/02-helm-install.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# monopod:tag:functional
+
+set -o errexit -o nounset -o errtrace -o pipefail -x
+
+IMAGE_DIR="$(basename "$(cd "$(dirname ${BASH_SOURCE[0]})/.." && pwd )")"
+IMAGE_NAME=${IMAGE_NAME:-"cgr.dev/chainguard/${IMAGE_DIR}:latest"}
```
found-it, line 8: Do we want to make the default for functional tests registry.local, that way we don't end up inadvertently testing an already published image if IMAGE_NAME is unavailable for some reason?

found-it:
This is awesome @jdolitsky!

images/fluent-bit/tests/02-helm-install.sh
jdolitsky: done!

found-it:
@jdolitsky do you want to merge this with a broken test? Or do you think we should fix the fluent-bit test first? I looked at it quickly last night and it seems like it's probably a file permission / missing plugin error/

ADDING_IMAGES.md
```diff
@@ -198,6 +198,18 @@ Also, be sure to make the script(s) executable:
 chmod +x images/superimg/test.sh
 ```
 
+## Functional testing
+
+In the same scripts as the section above, you may place the following comment
+somewhere at the top of the script:
+
+```
+# monopod:tag:functional
```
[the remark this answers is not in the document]
jdolitsky: howabout instead of "functional" these are "k8s" tests. like
```
# monopod:tag:k8s
```

kaniini (approved):

PR #376, opened 2023-03-24
Cleanup: Drop the extra config, `cgr.dev` is not default

mattmoor:
:broom: This cleans up the extra bit of config we had now that `cgr.dev` is the default domain in `chainctl`.

/kind cleanup

### Pre-review Checklist

#### For new image PRs only

- [ ] Image is marked experimental or stable as appropriate
- [ ] The last two minor versions are available
- [ ] ENTRYPOINT
- [ ] CMD:
- [ ] Add annotations e.g:
```
annotations:
```
- [ ] Include tests, at the very least a basic smoke test.

### DIFF
--- a/.github/actions/release-image/action.yml
+++ b/.github/actions/release-image/action.yml
@@ -291,16 +291,10 @@ runs:
         scanner: grype
         docker-config: .docker-rumble
 
-    - name: Setup chainctl
-      uses: chainguard-dev/actions/setup-chainctl@main
+    - uses: chainguard-dev/actions/setup-chainctl@main
       if: inputs.chainguardIdentity != ''
       with:
         identity: ${{ inputs.chainguardIdentity }}
-    - if: inputs.chainguardIdentity != ''
-      shell: bash
-      run: |
-        chainctl config set platform.registry https://cgr.dev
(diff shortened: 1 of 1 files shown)


.github/actions/release-image/action.yml
```diff
@@ -291,16 +291,9 @@ runs:
         scanner: grype
         docker-config: .docker-rumble
 
-    - name: Setup chainctl
-      uses: chainguard-dev/actions/setup-chainctl@main
-      if: inputs.chainguardIdentity != ''
+    - uses: chainguard-dev/actions/setup-chainctl@main
```
jdolitsky, line 294: still needs `if: inputs.chainguardIdentity != ''` since this being used in other places

mattmoor: Not just the two?

mattmoor: I added it back, but we should sort out what changes we need to make this our default path.

jdolitsky: it is at least also used here: https://github.com/wolfi-dev/tools/blob/main/.github/workflows/release.yaml#L86

jdolitsky (approved):

PR #412, opened 2023-03-31
Update static/README.md

imjasonh:
Remove the note about enterprise support, SLAs and older versions for `static`, which does not have older versions.


jdolitsky:
closing in favor of #413

Issue #246, opened 2023-01-27
Rust image missing rustup and config


amouat:
The rust image should include rustup so that users can add new targets (e.g. compile for other architectures) https://github.com/rust-lang/rustup

Also there seems to a config file missing e.g:

```
rustc config
error: couldn't read config: No such file or directory (os error 2)

error: aborting due to previous error
```

kaniini:
There should be a separate image for rust + rustup.  Using rustup to manage the rust toolchain will result in coverage gaps in SBOMs.

amouat:
Yeah, in retrospect I agree. I actually managed to work around the issue I had by using different flags. In general, rustup lets you add new versions of the compiler targetting different archs and systems, which is arguably something that should be provided by a separate image. We're not going to be able to cover every variation and use case however, so a rustup image is useful as an escape hatch.

patflynn:
@amouat how do you want to prioritise this?

amouat:
I'm wondering if we should put this in the dev image variant.

P2 for moment.

PR #420, opened 2023-04-03
Image: Add dev variant with rustup to rust image

patflynn:
Fixes:
https://github.com/chainguard-images/images/issues/246


jdolitsky (commented):
needs rebase, and then to run
```
(cd monopod && go install)
monopod readme
```

PR #426, opened 2023-04-05
Require IMAGE_NAME in all tests, do not default to public image

jdolitsky:


images/apko/tests/01-version.sh
```diff
@@ -2,7 +2,9 @@
 
 set -o errexit -o nounset -o errtrace -o pipefail -x
 
-IMAGE_DIR="$(basename "$(cd "$(dirname ${BASH_SOURCE[0]})/.." && pwd )")"
-IMAGE_NAME=${IMAGE_NAME:-"cgr.dev/chainguard/${IMAGE_DIR}:latest"}
+if [[ "${IMAGE_NAME}" == "" ]]; then
```
imjasonh, line 5: Won't nounset have the same effect with less boilerplate? We could just remove the default I think right?

jdolitsky: true, but unset wont crash until the var is used so other things may occur in the script beforehand)

feel free to suggest some other form of this

imjasonh (approved):

PR #427, opened 2023-04-05
Do not run k8s tests on main push

jdolitsky:


imjasonh:
Why not? 🤔

jdolitsky:
> Why not?

Well, for starters they are currently failing (and we know it). So everytime we merge something, this check runs and creates a red X on the latest commit on the repo.

dlorenc:
I did a bunch of debugging on why the test is failing and had the start of a fix - is there an issue open for that? i can dump my notes in there.

TLDR: The fluent-bit image we have works fine by itself, but when you deploy it inside the helm chart the helm chart overrides the default configuration at runtime, and the injected config doesn't work with our build configuration because we don't include all the optional dependencies.

jdolitsky:
@dlorenc opened https://github.com/chainguard-images/images/issues/431

PR #438, opened 2023-04-08
Add a test that the opensearch server starts and serves the default e…

dlorenc:
…ndpoint.

Fixes:

Related:


images/opensearch/tests/02-server.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
jdolitsky, line 1: this file needs to be set to executable

jdolitsky (approved):

jdolitsky (changes requested):
somehow need to get opensearch to bind to 0.0.0.0:9200 vs. 127.0.0.1:9200, this seems to be the problem with the test now

dlorenc:
> somehow need to get opensearch to bind to 0.0.0.0:9200 vs. 127.0.0.1:9200, this seems to be the problem with the test now

This is blocked on this: https://github.com/wolfi-dev/os/pull/1107

dlorenc:
> appears to still fail with newer package:
> 
> ```
> installing opensearch-2 (2.6.0-r2)
> ```

I had to increase the sleep to 20s and now it works.

jdolitsky (approved):

PR #454, opened 2023-04-13
run k8s tests on presubmit if label applied

jdolitsky:


found-it:
@jdolitsky I don't think it's in scope for this PR but I noticed the fluent-bit repo/tag parsing was a bit off in that failing test. 

https://github.com/chainguard-images/images/actions/runs/4693234209/jobs/8319967526?pr=454#step:8:29

Do you think it would be a good idea to use monopod or the action itself to inject the repo and tag so we remove the burden from users?

jdolitsky:
bleghhh .. i think it should go in the bash. PR incoming

luhring (approved):

PR #468, opened 2023-04-17
Updated ruby entrypoint and command

erikaheidi:
This PR updates the entrypoint for Ruby images to use `/usr/bin/ruby` with the default command set to `--version`.

The current entry point is set to the interactive ruby tool `irb`, resulting in odd behavior when you try to run Ruby scripts directly via `docker run`. This change was discussed in an [internal Slack thread](https://chainguard-dev.slack.com/archives/C0322APD4LU/p1681480063454969).

With the update, it is possible to execute a Ruby script with:

```shell
docker run --rm -v ${PWD}:/work cgr.dev/chainguard/ruby:latest script.rb
```

### DIFF
--- a/images/ruby/configs/3.0.apko.yaml
+++ b/images/ruby/configs/3.0.apko.yaml
@@ -25,7 +25,9 @@ paths:
 
 work-dir: /work
 
-cmd: /usr/bin/irb
+entrypoint:
+  command: /usr/bin/ruby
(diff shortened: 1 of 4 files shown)


erikaheidi:
The tests need to be updated to, working on it now.

amdawson (approved):

Issue #452, opened 2023-04-13
Nginx doesn't run on Kubernetes


amouat:
### Which image/versions are related to this issue/feature request?

cgr.dev/chainguard/nginx

### Issue/Feature description

The nginx image should run on Kubernetes, but it has issues binding to port 80:

```
$ kubectl run nginx --image cgr.dev/chainguard/nginx
pod/nginx created
$ kubectl get pods
NAME    READY   STATUS             RESTARTS      AGE
nginx   0/1     CrashLoopBackOff   2 (19s ago)   40s
$ kubectl logs nginx
2023/04/13 15:43:16 [emerg] 1#1: bind() to 0.0.0.0:80 failed (13: Permission denied)
```

mathieu-benoit:
FYI @amouat, some updates with ``:

This following flow works successfully:
```
kubectl run nginx --image cgr.dev/chainguard/nginx:next
kubectl expose pod nginx --port 80 --target-port 8080 --type LoadBalancer
```

But I'm still getting an error with:
```
cat <<EOF > nginx-next-unprivileged.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      automountServiceAccountToken: false
      securityContext:
        fsGroup: 65532
        runAsGroup: 65532
        runAsNonRoot: true
        runAsUser: 65532
        seccompProfile:
          type: RuntimeDefault
      containers:
        - name: nginx
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
                - ALL
            privileged: false
            readOnlyRootFilesystem: true
          image: cgr.dev/chainguard/nginx:next
          ports:
            - containerPort: 8080
          volumeMounts:
          - mountPath: /var/lib/nginx/tmp
            name: tmp
          - mountPath: /var/run
            name: run
      volumes:
      - emptyDir: {}
        name: tmp
      - emptyDir: {}
        name: run
EOF
kubectl expose deployment nginx --port=80 --target-port=8000 ----type LoadBalancer
```
In the logs I'm not seeing any obvious error:
```
nginx: [alert] could not open error log file: open() "/var/lib/nginx/logs/error.log" failed (30: Read-only file system)
[...]

Issue #288, opened 2023-02-21
Nginx Read Only File System


henrysachs:
A lot of Security people recommend running Images in Kubernetes or other Hypervisors in a read-only mode. [one example](https://docs.bridgecrew.io/docs/bc_k8s_21)
When trying to run the nginx image in read only mode like this: 

```
docker run -it -p8080:8080 --read-only --tmpfs /tmp cgr.dev/chainguard/nginx:latest
```

I receive the following error:

```
s6-svscan: fatal: unable to mkdir .s6-svscan: Read-only file system
```

tuananh:
related ? https://github.com/chainguard-dev/apko/issues/497

jdolitsky:
> related ? [chainguard-dev/apko#497](https://github.com/chainguard-dev/apko/issues/497)

This should not be related since we are using a pinned version of apko before that issue (497)

This may be able to be solved by creating this directory ahead of time in apko yaml, unsure about the inner workings of s6

kaniini:
Use `docker run -it -p8080:8080 --read-only --tmpfs /tmp --tmpfs /sv cgr.dev/chainguard/nginx:latest` to allow the supervisor to manage its state.

mathieu-benoit:
Same issue here.

mathieu-benoit:
Wondering if there is anything to reuse from this section below from the [`nginx-unprivileged` container image](https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/Dockerfile-alpine-slim.template#L109):
```
# implement changes required to run NGINX as an unprivileged user	
RUN sed -i 's,listen       80;,listen       8080;,' /etc/nginx/conf.d/default.conf \	
    && sed -i '/user  nginx;/d' /etc/nginx/nginx.conf \	
    && sed -i 's,/var/run/nginx.pid,/tmp/nginx.pid,' /etc/nginx/nginx.conf \	
    && sed -i "/^http {/a \    proxy_temp_path /tmp/proxy_temp;\n    client_body_temp_path /tmp/client_temp;\n    fastcgi_temp_path /tmp/fastcgi_temp;\n    uwsgi_temp_path /tmp/uwsgi_temp;\n    scgi_temp_path /tmp/scgi_temp;\n" /etc/nginx/nginx.conf \	
# nginx user must own the cache and etc directory to write cache and tweak the nginx config	
    && chown -R $UID:0 /var/cache/nginx \	
    && chmod -R g+w /var/cache/nginx \	
[...]

PR #478, opened 2023-04-19
Bring nginx closer to Docker official image.

amouat:
*THIS INCLUDES BREAKING CHANGES*

Most notably default port is now 8080 rather than 80, but also other config changes including default html directory and auto determining how many worker processes to instantiate.

This is all referenced in the README, which also includes information on how to get the old version. Is this enough or do we need to handle this differently?

Fixes: #452 #382

### DIFF
--- a/images/nginx/README.md
+++ b/images/nginx/README.md
@@ -33,7 +33,7 @@ If you are unable to update currently, you can use the last build of the previou
 docker pull cgr.dev/chainguard/nginx@sha256:bcc6b0d052298112e4644b258de0fa4dc1509e3df8f7c0fba09e8c92987825e7
 ```
 
-This digest corresponds to nginx:1.24.0. This image is not updated and you should migrate to the
+This digest corresponds to nginx version 1.24.0. This image is not updated and you should migrate to the
 new configuration as soon as possible.
 
 ## Get It!
--- a/images/nginx/README.md
+++ b/images/nginx/README.md
@@ -98,7 +98,7 @@ or an equivalent.
 ## Differences to [Docker Official Image](https://hub.docker.com/_/nginx)
 
 Where possible, the image tries to stay close to Docker official image configuration. However our
-image strives for minimalism and the default image does not include a shell or pacakge manager,
+image strives for minimalism and the default image does not include a shell or package manager,
 meaning it isn't possible to achieve an identical configuration. In this section we outline the
 major differences.
 
--- a/images/nginx/README.md
+++ b/images/nginx/README.md
@@ -13,12 +13,12 @@
 
 A minimal nginx base image rebuilt every night from source.
 
-## Upcoming Breaking Changes
+## Breaking Changes
 
-On May 3 2023 the Chainguard nginx image will be rebuilt with several improvements, including
-breaking changes. You may need to take action to update your application to prevent disruption. 
+On May 3 2023 the Chainguard nginx image was rebuilt with several improvements, including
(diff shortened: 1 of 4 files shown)


patflynn:
This looks great.  @eddiezane @kaniini for approval.  

Let's quickly discuss offline the consequence of the breaking change.

amdawson:
OK I guess let's go for it

amdawson (approved):

mathieu-benoit:
Wow, thank you for the hard work on this @amouat! Looking forward to testing it!

Out of curiosity, is it fixing https://github.com/chainguard-images/images/issues/288 too?

amouat:
I have now made a mess of the git history somehow :sob:

amouat:
Now a test is failing :sob:

images/nginx/README.md
```diff
@@ -13,12 +13,12 @@
 
 A minimal nginx base image rebuilt every night from source.
 
-## Upcoming Breaking Changes
+## Breaking Changes
 
-On May 3 2023 the Chainguard nginx image will be rebuilt with several improvements, including
-breaking changes. You may need to take action to update your application to prevent disruption. 
+On May 3 2023 the Chainguard nginx image was rebuilt with several improvements, including
```
imjasonh, line 18: Can you link an issue/PR to describe why this change was made, or what kinds of changes were made?

images/nginx/README.md
```diff
@@ -13,12 +13,12 @@
...
+breaking changes. You may need to take action to update your application. 
 
-Specifically, the config file is being changed to bring the default configuration closer to that of
+Specifically, the config file was changed to bring the default configuration closer to that of
 official nginx image. If you override the config with a custom configuration, you should not be affected.
 
 The changes include:
```
imjasonh, line 24:
```suggestion
The changes included:
```

images/nginx/README.md
```diff
@@ -62,3 +67,70 @@ To use a custom `nginx.conf` you can mount the file into the container
...
+
+The warning is telling us container is already running as the named user, so it doesn't have
+anything to do. If the container is run as root, it would switch to the named user. We decided to
+leave this configuration in despite the warning for anyone that runs with `--user` switch in Docker
+or an equivalent.
+
+## Differences to Docker Official Image
+
+Where possible, the image tries to stay close to Docker official image configuration. However our
+image strives for minimalism and the default image does not include a shell or pacakge manager,
```
imjasonh, line 101:
```suggestion
image strives for minimalism and the default image does not include a shell or package manager,
```

images/nginx/README.md
```diff
@@ -62,3 +67,70 @@ To use a custom `nginx.conf` you can mount the file into the container
...
+```
+nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
+```
+
+The warning is telling us container is already running as the named user, so it doesn't have
+anything to do. If the container is run as root, it would switch to the named user. We decided to
+leave this configuration in despite the warning for anyone that runs with `--user` switch in Docker
+or an equivalent.
+
+## Differences to Docker Official Image
```
imjasonh, line 98:
```suggestion
## Differences to [Docker Official Image](https://hub.docker.com/_/nginx)
```

imjasonh (changes requested):

imjasonh (approved):

Issue #407, opened 2023-03-30
[Image Request]: clang


AnishShah:
### What software would you like us to add to our image catalog? (one per issue please)

Filing this bug to request an image for `clang`.

PR #491, opened 2023-04-21
Image: Add clang 15 image. 

patflynn:
Fixes: #407

### Pre-review Checklist

- [x] **IMPORTANT: 'image-request' tag has been applied if this PR is adding any images, including new versions or variants**

#### For new image PRs only

- [x] Image is marked experimental or stable as appropriate
- [x] The version included is the latest GA version of the software
- [x] The latest tag points to the newest stable version
- [x] ENTRYPOINT
  - [x] For applications/servers/utilities call main program with no arguments e.g. [redis-server]
- [x] CMD:
  - [x] For utilities/tooling bring up help e.g. `–help`
- [ ] Add annotations e.g:
```
annotations:
```
- [x] Include tests, at the very least a basic smoke test.


patflynn:
@kaniini assigned to you for review because I have no idea what I'm doing.  

One question: is running as root the right thing here? or is there a better way?

jdolitsky (commented):
do we want `clang-15:latest` or `clang:15`/ `clang:latest` ? This would be first image with a version number in it

patflynn:
@jdolitsky  @kaniini  good question. I don't know much about that tool but since it's clang-15 in wolfi I assumed the version stream should be locked to the major.

patflynn:
WRT to the 15 tag, I don't want to add that for now.

jdolitsky:
postgres is a good example. image name just "postgres:latest", but we built using postgresql-15: https://github.com/chainguard-images/images/blob/main/images/postgres/configs/latest.apko.yaml#L11

jdolitsky (approved):

PR #498, opened 2023-04-22
The k8s tests seem to have been changed to separate the image registr…

dlorenc:
…y and repo.

Update influx and weviate to match the new pattern.


jdolitsky (approved):
looks like weaviate tests were already failing, so approved

PR #511, opened 2023-04-26
add PATH env var to all images, format with yam

imjasonh:
We've been very inconsistent about adding `PATH` env vars to our images, meaning we can't run images like this:

```
$ docker run --entrypoint sh cgr.dev/chainguard/go:latest-dev go version                      
sh: can't open 'go': No such file or directory
```

Instead you have to specify `/usr/bin/go` because the `PATH` env var is unset.

To fix this, this change was done for all configs, using `yq`:

```
for f in images/**/*.apko.yaml; do yq -i '.environment.PATH = "/usr/sbin:/sbin:/usr/bin:/bin"' $f; done
```

This adds `.environment.PATH` for all files, or updates it if it exists. I spot-checked and reverted some (like [rabbitmq](https://github.com/chainguard-images/images/blob/main/images/rabbitmq/configs/latest.apko.yaml#L59)) that already had a custom PATH.

This also formats all the files with [`yam`](https://github.com/chainguard-dev/yam/), with a yam config added to the repo. This PR also adds a lint check in CI.

⚠️ We should add a lint check that `PATH` is set in new images, or else we're going to have another PR like this again in a few months ⚠️


images/deno/configs/latest.apko.yaml
```diff
@@ -19,10 +19,11 @@ paths:
 work-dir: /app
 
 environment:
-  PATH: /usr/sbin:/sbin:/usr/bin:/bin
```
imjasonh, line 22: Should we include `/sbin` in the PATH too?

jdolitsky (approved):

images/deno/configs/latest.apko.yaml
imjasonh: I think the answer will be **yes**, because a couple packages install themselves there: https://github.com/search?q=repo%3Awolfi-dev%2Fos%20%24%7B%7Btargets.subpkgdir%7D%7D%2Fsbin&type=code

images/apko/configs/latest.apko.yaml
```diff
@@ -9,18 +9,16 @@ contents:
     - alpine-keys
     - apko
     - wolfi-baselayout
-
```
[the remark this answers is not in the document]
imjasonh: I think I'd accidentally reverted the yam-formatting. Re-running `yam images/**/*.yaml` fixed it.

luhring (approved):

images/apko/examples/wolfi-base.yaml
```diff
@@ -6,7 +6,6 @@ contents:
     - wolfi-baselayout
 
 cmd: /bin/sh -l
-  
-# optional environment configuration
+
 environment:
-  PATH: /usr/sbin:/sbin:/usr/bin:/bin
+  PATH: /usr/sbin:/usr/bin:/bin
```
amouat, line 11: why is `:/sbin` removed?

images/rabbitmq/configs/latest.apko.yaml
```diff
@@ -56,7 +56,7 @@ environment:
   RABBITMQ_CONF_ENV_FILE: /etc/rabbitmq/rabbitmq-env.conf
   HOME: /var/lib/rabbitmq
   LANG: en_US.UTF-8
-  PATH: /usr/sbin:/sbin:/usr/bin:/bin:/usr/lib/rabbitmq/bin/
+  PATH: /usr/sbin:/sbin:/usr/bin:/bin
```
amouat, line 59: Check this one!

amouat (commented):
Think this is merged already, but please double check rabbitmq and wolfi-base.

PR #514, opened 2023-04-27
Fix rabbitmq path

amouat:
Rabbitmq lib dir was accidentally removed  during mass path fix.

Related: https://github.com/chainguard-images/images/pull/511


amouat:
Hmm, I've made a mess of this git history 🤦

PR #516, opened 2023-04-27
Revert "add PATH env var to all images, format with yam"

imjasonh:
Reverts chainguard-images/images#511

The justification for the change was flawed.

This doesn't work:
```
$ docker run --entrypoint sh cgr.dev/chainguard/go:latest-dev go version
sh: can't open 'go': No such file or directory
```

but this does:

```
$ docker run --entrypoint sh cgr.dev/chainguard/go:latest-dev -c "go version"
go version go1.20.3 linux/arm64
```

This seems to be because the shell (busybox in this case) provides a default `PATH` if none is set:

```
$ docker run --entrypoint sh cgr.dev/chainguard/go:latest-20230301 -c "echo \$PATH"
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

This is with an older version before #511; after #511 this is changed to:

```
$ docker run --entrypoint sh cgr.dev/chainguard/go:latest-20230427 -c "echo \$PATH"
/usr/sbin:/sbin:/usr/bin:/bin
```

Since the change wasn't necessary, and actually removed and reordered items from the `PATH` which folks may have depended on, we should revert it.


jdolitsky (approved):

luhring:
Just so I'm tracking — if I wanted to use the go image to invoke shell commands, I'd need to remember to put `-c` first?

found-it:
Should we capture this as a "Lesson Learned - Do not do" in the best practices doc?

imjasonh:
> Just so I'm tracking — if I wanted to use the go image to invoke shell commands, I'd need to remember to put `-c` first?

...and `--entrypoint sh` to tell it to run the shell, which provides the PATH.

Issue #513, opened 2023-04-27
PostgreSQL encoding is `SQL_ASCII` instead of `UTF-8` 


wilsonehusin:
### Which image/versions are related to this issue/feature request?

cgr.dev/chainguard/postgres@sha256:405031c3608549b5aed1069774ba7072b1ff8e2d0a42da648bec56c29ff9a10f

### Issue/Feature description

I swapped an existing usage of Docker library's Postgres image and expecting things to mostly work, but somehow encoding isn't playing along. Then I found out that Docker library's PostgreSQL image uses `UTF-8` encoding out of the box, but Chainguard's is on `SQL_ASCII`.

I'd love for the defaults to match and support the experience of just swapping images, but at the very least some instructions on getting UTF-8 or migrating from Docker library's image would be appreciated!

wilsonehusin:
Looks like setting `LC_ALL=en_US.UTF-8` did the trick!

dlorenc:
We should be able to default to this!

imjasonh:
Does someone want to send a PR to set this default? If so assign me and I'll approve it.

wilsonehusin:
> Does someone want to send a PR to set this default? If so assign me and I'll approve it.

@imjasonh I put up #523 with my best understanding of how contributions on this project works. Let me know if there is anything else I can do to conform with the project's convention!

PR #523, opened 2023-04-30
Fix: use UTF-8 encoding on Postgres image by default

wilsonehusin:
Postgres (the program) honors environment locale, which would be SQL_ASCII when unset. Docker library builds their image with UTF-8 by default[^0]. This change is to ensure users have a closer experience between Docker images and Chainguard images.

Added test as well!

[^0]: https://github.com/docker-library/postgres/blob/887333be0892981ffbf468fa62aa6f2e0dd6358a/15/alpine/Dockerfile#L19-L21

Fixes: #513


images/postgres/tests/03-locale.sh
```diff
@@ -0,0 +1,15 @@
...
+    echo "Must set IMAGE_NAME environment variable. Exiting."
+    exit 1
+fi
+
+CONTAINER_NAME=${CONTAINER_NAME:-"postgres-$(date +%s)"}
+
+docker run -e POSTGRES_PASSWORD=password -d --name $CONTAINER_NAME $IMAGE_NAME
+trap "docker logs $CONTAINER_NAME && docker rm -f $CONTAINER_NAME" EXIT
+sleep 5
+docker exec -it $CONTAINER_NAME /usr/bin/psql -U postgres -t --csv -c "show server_encoding;" | head -n 1 | grep UTF8
```
jdolitsky, line 15:
```suggestion
docker exec $CONTAINER_NAME /usr/bin/psql -U postgres -t --csv -c "show server_encoding;" | head -n 1 | grep UTF8
```

jdolitsky (approved):
thank you!

PR #524, opened 2023-05-01
Modified permissions on directories to support tmpfs.

amouat:
We want to be able to run nginx with:

```
docker run --read-only --tmpfs /var/lib/nginx/tmp/ --tmpfs /var/run -p 8080:8080 cgr.dev/chainguard/nginx:next
```

But it currently fails with permission errors.

The problem seems to be that `--tmpfs` does not set permissions correctly even when run with the `tmpfs-mode` option e.g: `--mount type=tmpfs,destination=/var/lib/nginx/tmp/,tmpfs-mode=777`. Probably related to https://github.com/moby/moby/issues/40881.

Related: https://github.com/chainguard-images/images/issues/288

This is *not* a breaking change (permissions are being relaxed).

### DIFF
--- a/images/nginx/configs/next.apko.yaml
+++ b/images/nginx/configs/next.apko.yaml
@@ -30,13 +30,15 @@ paths:
     uid: 65532
     gid: 65532
     type: directory
-    permissions: 0o755
+    # Wide permissions required for running with tmpfs. Seems to be related to Docker bug https://github.com/moby/moby/issues/40881
+    permissions: 0o777
     recursive: true
   - path: /var/run
     uid: 65532
     gid: 65532
     type: directory
-    permissions: 0o755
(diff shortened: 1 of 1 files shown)


amouat:
Wait a moment, I think there might still be an issue with the log files...

amouat:
So I was worried about:

```
nginx: [alert] could not open error log file: open() "/var/lib/nginx/logs/error.log" failed (30: Read-only file system)
```

But it turns out this message is printed out because nginx tries to write to this log-file before even opening the config file. It's just a warning and doesn't affect anything.

More information here:
https://stackoverflow.com/questions/34258894/nginx-still-try-to-open-default-error-log-file-even-though-i-set-nginx-config-fi

That being said, I couldn't figure out how the official nginx unprivileged image avoids this warning.

I don't think it's worth exploring this rabbit-hole further, let's just document it.

jdolitsky (approved):

PR #531, opened 2023-05-02
Feature: Add signing policy to improve SLSA

patflynn:
I'm guessing I also need to setup the CG app somehow. 

What other OIDC providers should we allow?

### DIFF
--- a/.chainguard/source.yaml
+++ b/.chainguard/source.yaml
@@ -1,0 +1,13 @@
+# Copyright 2022 Chainguard, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+spec:
+  authorities:
+    - keyless:
+        # allow commits signed by users using GitHub or Google OIDC
+        identities:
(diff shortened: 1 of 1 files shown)


imjasonh:
We should add something to [ADDING_IMAGES](https://github.com/chainguard-images/images/blob/main/ADDING_IMAGES.md) to say that commit signing is required, and why. I can draft something up.

imjasonh (dismissed):
I'm for this change. I'm going to put a hold on the PR so we can solicit feedback from folks.

strongjz:
We should allow all the providers that Sigstore does, so we can still use contributions from the community, specifically GitHub since they need that identity to make PRs.

strongjz:
Google and GitHub are a great start, I see no reason why we should not keep it open to all the providers that Sigstore supports.

kaniini:
LGTM

kaniini (approved):

PR #538, opened 2023-05-03
Tweak the zookeeper environment variables and path and add a test

dlorenc:
Fixes:

Related:


patflynn (approved):
should we remove the 'experimental' label while we're here? @imjasonh

images/zookeeper/configs/latest.apko.yaml
```diff
@@ -11,6 +11,7 @@ contents:
     - bash
     - zookeeper
     - openjdk-17-jre
+    - vim
```
jdolitsky, line 14: assuming this was just for testing purposes?
```suggestion
```

PR #555, opened 2023-05-08
Fix oauth2-proxy test 01-version.sh

jdolitsky:


dlorenc (approved):

jdolitsky:
likely related to https://github.com/wolfi-dev/os/pull/1880

jdolitsky:
the path changes to /usr/bin, will update this PR

images/oauth2-proxy/configs/latest.apko.yaml
```diff
@@ -19,7 +19,7 @@ accounts:
   run-as: oauth2-proxy
 
 entrypoint:
-  command: /bin/oauth2-proxy
+  command: /usr/bin/oauth2-proxy
```
jdolitsky, line 22: @found-it - any issue with this?

dlorenc: Looks like this was my change when I moved this to the standard go/build pipeline: https://github.com/wolfi-dev/os/blob/8259eca08a08507733990a2c94e0b3d6d6425d8d/oauth2-proxy.yaml#L32

/usr/bin is our standard here for these, fwiw

found-it: Yep as long as `/usr/bin` is the standard (just separately confirmed it's where the oauth2-proxy binary should be) then there is no issue @jdolitsky

PR #562, opened 2023-05-09
explicitly re-add default envs for images that overrode them

imjasonh:
⚠️ Please review this carefully ⚠️ 

This was an action item from the `PATH` change post-mortem from a couple weeks ago.

`apko` sets default environment variables when the config doesn't specify any. Specifically, it sets:

- `PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
- `SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt`

([here](https://github.com/chainguard-dev/apko/blob/87f7f0ff54fab41ca6dd580343487310e8e25d94/pkg/build/oci/oci.go#L204-L209))

Crucially, it only sets these when the apko config YAML doesn't specify any envs -- if it specifies any envs, those default envs are not added.

This means that, for example, `go`'s apko.yaml setting `GODEBUG` by default, also _unsets_ `PATH` and `SSL_CERT_FILE`. Presumably this behavior is not catastrophic, since we haven't heard any complaints about this difference between the 1.19 and 1.20 images. But, it's still surprising, and _could_ break someone coming from an image that has no env var set, where they get the implicit default, to an image that has another unrelated env var set, where the implicit defaults are missing.

Note also that rabbitmq has a different `PATH` than this default, so its value should be retained, and not reverted to apko's implicit default.

Some configs specified `PATH` that was just a subset of the implicit path, and I believe these should be avoided -- the default `PATH` is more consistent and less surprising. _Please speak up if you believe some images should have different `PATH`s that are a subset of the default._

In general we might want to be totally explicit about the envs in all configs, enforced by linting. We may also want to stage this change out, though none of the images affected by this change are in the "high profile" category (static, busybox, git, wolfi-base). We may even want to warn about `apko`'s implicit env behavior, and eventually remove it -- this may be a surprising breaking change for `apko` users.

I'm going to mark this PR as draft until I've gotten approvals from all of @amdawson @amouat @patflynn @kaniini

### DIFF
--- a/images/apko/examples/wolfi-base.yaml
+++ b/images/apko/examples/wolfi-base.yaml
@@ -6,7 +6,3 @@ contents:
     - wolfi-baselayout
 
 cmd: /bin/sh -l
-  
-# optional environment configuration
-environment:
-  PATH: /usr/sbin:/sbin:/usr/bin:/bin
--- a/images/bazel/configs/latest.apko.yaml
+++ b/images/bazel/configs/latest.apko.yaml
@@ -27,6 +27,8 @@ entrypoint:
 
 environment:
   JAVA_HOME: /usr/lib/jvm/openjdk
+  PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+  SSL_CERT_FILE: /etc/ssl/certs/ca-certificates.crt
 
 work-dir: /home/bazel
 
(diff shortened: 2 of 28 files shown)


mattmoor (changes requested):
to let folks review

tuananh:
@imjasonh or we change it in apko to only set `PATH` and `SSL_CERT_FILE` if that specific var is not yet set => more like merge with higher priority from the image env var

PR #569, opened 2023-05-10
remove dotnet-6 from public catalog

patflynn:
removing this image since we don't intend to support older versions in the free catalog.


kaniini (approved):

jdolitsky:
what about aspnet?

patflynn:
@jdolitsky good catch! https://github.com/chainguard-images/images/pull/570

PR #577, opened 2023-05-11
mark certmanager image as stable

joshrwolf:
we've been dogfooding cert-manager for a while now and we're comfortable upgrading this to stable 🎉


imjasonh (approved):
🚀

PR #580, opened 2023-05-11
Add prometheus operator

found-it:
Fixes:

Related: 

### Pre-review Checklist

#### For new image PRs only

- [x] Include tests, sufficient enough that you would trust this image running in production.
- [x] The version included is the latest GA version of the software
- [x] The latest tag points to the newest stable version
- [x] ENTRYPOINT
- [x] Consider where and how the image deviates from popular alternatives. Is there a good reason and is it documented?
- [x] Add annotations e.g:
```
annotations:
```
- [x] Check if environment variables are needed e.g. to set data locations
- [x] Ensure the image responds to SIGTERM
  - [x] `docker kill $(docker run -d --rm cgr.dev/chainguard/nginx)`
- [x] Documentation. Let's make this excellent. Include usage example.
- [x] Error logs write to stderr and normal logs to stdout. DO NOT write to file.


images/prometheus-operator/tests/01-smoke.sh
```diff
@@ -0,0 +1,120 @@
...
+  echo "${signal} Success"
+}
+
+#
+# Basic functional testing
+#
+TEST_startup
+TEST_default_args
+
+# TODO: Container won't start unless it's in a pod. Need to test on k8s
```
Dentrax, line 118: Can you please add a new `02-helm.sh` for Kubernetes testing? See: https://github.com/chainguard-images/images/blob/main/ADDING_IMAGES.md#kubernetes-testing

Here is an example: https://github.com/chainguard-images/images/blob/d98c6725d764d540df4d30ee08c0c74505f1e6ee/images/influxdb/tests/03-helm.sh#L26-L33

found-it: Just added @Dentrax!

joshrwolf (approved):

PR #587, opened 2023-05-12
Create alternate actions path for tf-apko, opt-in ko

jdolitsky:
Also: generate the SLSA provenance ahead of time and cache it to prevent rate limiting

Plz wait to merge - waiting on tf-apko release to fix issue related to x86_64 vs. amd64

### DIFF
--- a/.github/actions/release-image-terraform/action.yml
+++ b/.github/actions/release-image-terraform/action.yml
@@ -1,4 +1,4 @@
-name: release-image
+name: release-image-terraform
 inputs:
   slackWebhookUrl:
     default: ''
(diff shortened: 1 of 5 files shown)


.github/actions/release-image-terraform/action.yml
```diff
@@ -0,0 +1,554 @@
...
+        TF_VAR_apko_config_path: ${{ inputs.apkoConfig }}
+        TF_VAR_target_repository: ${{ inputs.apkoBaseTag }}
+        TF_VAR_extract_package: ${{ inputs.apkoPackageVersionTag }}
+        TF_VAR_build_options: ${{ inputs.apkoBuildOptions }}
+        DOCKER_CONFIG: .docker-rumble
+      run: |
+        set -x
+        env | grep '^TF_VAR_'
+        terraform init
+        terraform apply -auto-approve
```
jdolitsky, line 182: here's where we actually call terraform

main.tf
```diff
@@ -0,0 +1,54 @@
...
+
+output "image_ref" {
+  value = module.image.image_ref
+}
+
+output "package_version" {
+  value = [
+    for x in module.image.config.contents.packages : regexall("(((([a-z0-9]+)(?:[.][a-z0-9]+)?)(?:[.][a-z0-9]+)?)(?:[-][a-z0-9]+)?)", trimprefix(x, "${var.extract_package}=")) if startswith(x, "${var.extract_package}=")
+  ][0][0]
+}
```
imjasonh, line 54: We have to move this into some kind of `apko_tag` resource, we can't live like this.

.github/actions/release-image-terraform/action.yml
jdolitsky: still need to add support for build options (to enable -dev variants etc)

main.tf
```diff
@@ -0,0 +1,54 @@
...
+
+variable "extract_package" {
+  description = "TODO."
+}
+
+provider "apko" {
+  extra_repositories = []
+  extra_keyring      = []
+  default_archs      = []
+  extra_packages     = []
```
imjasonh, line 36:
```suggestion
  extra_repositories = ['https://packages.wolfi.dev/os']
  extra_keyring      = ['https://packages.wolfi.dev/os/wolfi-signing.rsa.pub']
  default_archs      = ['x86_64', 'aarch64']
  extra_packages     = ['wolfi-baselayout']
```

Is it okay to make these changes, and remove the underlying config from ko's apko YAMLs? Or would you prefer that we wait for this to land as-is first? (I'm fine either way, so long as we can get this common config in here eventually)

imjasonh (approved):

main.tf
jdolitsky: im looking to land and see what breaks for now 😅

i will follow up shortly

main.tf
imjasonh: Filed https://github.com/chainguard-dev/terraform-publisher-apko/issues/10 to discuss and track

PR #592, opened 2023-05-13
Allow image-summary to operate on images with only signed SBOMs

jdolitsky:


.github/actions/image-summary/action.yml
```diff
@@ -61,9 +65,13 @@ runs:
...
             primary="$(echo "${line}" | jq -r .primary)"
             set -x # Try to debug any cosign failures here
+
             # Support single-arch images (for now)
-            cosign download sbom --platform=linux/amd64 "${ref}:${primary}" > sbom.json || \
-              cosign download sbom "${ref}:${primary}" > sbom.json
+            digest="$(crane digest --platform=linux/amd64 "${ref}:${primary}" || crane digest "${ref}:${primary}")"
+            cosign download attestation \
+              --predicate-type https://spdx.dev/Document \
+              "${ref}@${digest}" | jq -r .payload | base64 -d > sbom.json
```
mattmoor, line 73: Can you confirm that this returns the `predicate` and not the in-toto wrapper or DSSE or something where we need further extraction login?

mattmoor (approved):

.github/actions/image-summary/action.yml
jdolitsky: realllly good call. needs an extra `| jq -M -c .predicate`

PR #610, opened 2023-05-15
update README for dex

joshrwolf:
Updates dex docs to fix typos


amouat (approved):
Left a couple of suggestions.

Would it be worth starting with showing just a simple `docker run`? I know that's not how people use the image in reality, but it might be nice to get people started testing/debugging.

PR #617, opened 2023-05-16
use new JAVA_HOME path and *-default-jvm for jdk/jre-11

joshrwolf:
Brings 11 up to the new standard installation path for java, and uses the `-default-jvm` package.

Related: https://github.com/wolfi-dev/os/pull/2075

### DIFF
--- a/images/jdk/configs/openjdk-11.apko.yaml
+++ b/images/jdk/configs/openjdk-11.apko.yaml
@@ -9,6 +9,7 @@ contents:
     - glibc-locale-en
     - busybox
     - openjdk-11
+    - openjdk-11-default-jvm
     - libstdc++
 
 accounts:
@@ -25,7 +26,7 @@ work-dir: /home/build
 
 environment:
   LANG: en_US.UTF-8
-  JAVA_HOME: /usr/lib/jvm/openjdk
+  JAVA_HOME: /usr/lib/jvm/java-11-openjdk
 
 paths:
   - path: /home/build
--- a/images/jre/configs/openjdk-11.apko.yaml
+++ b/images/jre/configs/openjdk-11.apko.yaml
@@ -8,6 +8,7 @@ contents:
     - wolfi-baselayout
     - glibc-locale-en
     - openjdk-11-jre
+    - openjdk-11-default-jvm
     - libstdc++
 
 accounts:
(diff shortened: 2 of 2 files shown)


joshrwolf:
draft until jdk11 packages are updated

images/jre/configs/openjdk-11.apko.yaml
```diff
@@ -19,14 +20,14 @@ accounts:
       uid: 65532
   run-as: java
 
-work-dir: /app
+work-dir: /home/build
```
patflynn, line 23: could this be breaking?

patflynn:
the package is updated already isn't it? https://github.com/wolfi-dev/os/pull/2075

images/jre/configs/openjdk-11.apko.yaml
joshrwolf: potentially yeah, if folks are using absolute paths within their build environment. do we have precedence for these kind of changes?

joshrwolf:
> the package is updated already isn't it? [wolfi-dev/os#2075](https://github.com/wolfi-dev/os/pull/2075)

it was missing aarch64, which should be g2g now, bringing this back as ready

images/jre/configs/openjdk-11.apko.yaml
patflynn: I'll defer to @imjasonh on making a call on this.

patflynn: maybe move to a different PR or is this related to the jdk refactoring?

kaniini: Not related.  `/home/build` is just the default build workspace in melange, but does not matter here, except for scenarios where we might want it to be consistent for education reasons.

images/jdk/configs/openjdk-11.apko.yaml
```diff
@@ -9,6 +9,7 @@ contents:
     - glibc-locale-en
     - busybox
     - openjdk-11
```
patflynn, line 11: based on what I heard from @kaniini we just need the openjdk-11-default-jvm package as it pulls in openjdk-11 as a runtime dep.

patflynn: actually I think I misunderstood. hold off on that change

patflynn: I think my understanding now is that -default-jvm only pulls in the JRE and not the JDK. So perhaps we keep it here but remove from the JRE image?  I guess it would have no impact either way because it's a subset or equivalent?

patflynn (approved):

PR #621, opened 2023-05-17
{jdk,jre}: fix openjdk 11 variants to use -default-jvm

kaniini:


images/jre/configs/latest.apko.yaml
```diff
@@ -26,7 +26,7 @@ entrypoint:
 
 environment:
   LANG: en_US.UTF-8
-  JAVA_HOME: /usr/lib/jvm/openjdk
+  JAVA_HOME: /usr/lib/jvm/java-11-openjdk
```
patflynn, line 29: doesn't latest stay on the 17 JDK path for now?

kaniini: oh oops

PR #635, opened 2023-05-18
Update images digests

chainguardian:
Update images digests

```release-note
NONE
```


mattmoor:
cc @kaniini @imjasonh 

When we're ready to canary the `kontext` changes they should be in these build IIUC 🤞

mattmoor (approved):

PR #659, opened 2023-05-23
Fix regex in test to capture any version string (it previously used t…

dlorenc:
…o say canary, but now the yaml is v3.5.0).


dlorenc:
This passed here: https://github.com/dlorenc/images/actions/runs/5055616407/jobs/9072056285

rawlingsj (approved):

PR #662, opened 2023-05-23
This drops the policy check for vulnz attestations

mattmoor:
Currently we publish vuln attestations whenever we publish a new image, but we never refresh these attestations, so they are of limited use to anyone consume them because they virtually never show anything when they are produced and they are never refreshed, so they are basically a test of staleness tolerance :)

I'm going to see if I can remove the attesting while I'm at it, but staging the policy check removal first while I track that down.


amdawson:
We do sell this as a feature. Need to be cautious about removing it for any length of time

mattmoor:
@amdawson this doesn't actually check anything meaningful for what we sell.

What we do today is akin to getting a covid test at birth and forever claiming you have a negative test.  We should be testing every container still getting pulls ~daily, which is not what this is at all. If that is what we want (I think it is), then we should prioritize it for Q3.

As it stands, I would be shocked if anyone was looking at these but us.

mattmoor:
To remove the attestation itself appears to involve dropping `-attest` here (and the related code): https://github.com/chainguard-dev/rumble/blob/4fcec3a34a79224872eee15e3a5e3b22ba52fa2f/action.yaml#L40

mattmoor:
We invoke that here, and we should keep it because this is populating rumble:
https://github.com/chainguard-images/images/blob/892c1dd7630c1ca86a79c38e97b534cbbafbf241/.github/actions/release-image/action.yml#L238-L363

amdawson:
@pdeslaur @jspeed-meyers please confirm you're not using the vuln attestation data on images for anything (like vulnz in UI or rumble)

pdeslaur:
No concerns on my side.

mattmoor:
@amdawson we will keep publishing the rumble stuff, just stop attesting.  IDK why those two things got conflated, but it's an unnecessary source of confusion.  The proposal is just: stop attesting (here), but keep invoking the rumble stuff.

mattmoor:
This change also just stops the policy check in images, we will need a change to rumble to actually stop attesting fwiw.

amdawson (approved):
Ok

jspeed-meyers:
@mattmoor and @amdawson, @jdolitsky is the true expert on `rumble`. FWIW, my own understanding matches @mattmoor's : the rumble data does NOT depend on the attesting. So, from my perspective, you're good to go :)

mattmoor:
I'm going to merge this and stage a PR to stop `-attest` in rumble.

PR #671, opened 2023-05-25
Remove notice about authentication in April

luhring:
Removes the message "Starting in April, accessing older tags will require authentication." from the `redis` and `php` images.


amdawson (approved):
approved, but should probably have general notice on images docs that auth will be required for versioned tags starting August 16, 2023

PR #674, opened 2023-05-25
Drop qemu from terraform-based image release

mattmoor:
@kaniini should we be able to remove this from the non-TF flow as well?


jdolitsky:
this appears to be outdated. closing

PR #677, opened 2023-05-25
add opa gatekeeper

developer-guy:
Fixes:

Related: https://github.com/wolfi-dev/os/pull/1983

### Pre-review Checklist

#### For new image PRs only

- [x] ENTRYPOINT
- [ ] CMD:
- [x] Add annotations e.g:
```
annotations:
```


amouat:
Nice test!

images/gatekeeper/configs/latest.apko.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - groupname: nonroot
+      gid: 65532
+  users:
+    - username: nonroot
+      uid: 65532
+      gid: 65532
+  run-as: 65532
+
+entrypoint:
+  command: manager
```
amouat, line 22: If this works, it's fine, but I'm surprised we don't need the full path?

Could also add a cmd with `--help` if it has one?

images/gatekeeper/README.md
```diff
@@ -0,0 +1,22 @@
...
+| **Status** | stable |
+| **OCI Reference** | `cgr.dev/chainguard/gatekeeper` |
+| **Variants/Tags** | ![](https://storage.googleapis.com/chainguard-images-build-outputs/summary/gatekeeper.svg) |
+
+*[Contact Chainguard](https://www.chainguard.dev/chainguard-images) for enterprise support, SLAs, and access to older tags.*
+
+---
+<!--monopod:end-->
+
+Minimalist Wolfi-based Gatekeeper which is a policy controller for Kubernetes
```
amouat, line 14: We need some usage docs, even if it just points to how to use it in a Helm a chart, or however it's normally run.

amouat (changes requested):
Looks good, but missing docs and it would be worth double-checking entrypoint & cmd.

images/gatekeeper/README.md
developer-guy: Thanks @amouat, I've added some usage details and updated the cmd with help command.

amouat (commented):
Could we add a short example of how to override the image in the upstream Helm chart?

Also could we do a one-liner using kustomize for the yaml install?

amouat (approved):

PR #680, opened 2023-05-26
Expose the "extra" settings as TF variables.

mattmoor:
This will let us pass different values for these across contexts as we continue to evolve these workflows

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -25,22 +25,41 @@ variable "target_repository" {
   description = "The docker repo into which the image and attestations should be published."
 }
 
+variable "extra_repositories" {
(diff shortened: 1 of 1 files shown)


cpanato (dismissed):

mattmoor:
I had to cut a release of the module, rerunning 🤞

mattmoor:
I added a module version pin, so it's clear what version of things we use.

mattmoor:
ok, I believe `ko` is finally passing, it took a little while for the module to show up (likely bad caching in the TF registry)

imjasonh (approved):

PR #682, opened 2023-05-26
hugo: add extended image variant

marshall007:

#### For new image PRs only

- [ ] ENTRYPOINT
- [ ] CMD:
- [ ] Add annotations e.g:
```
annotations:
```


johnfosborneiii (dismissed):

imjasonh (dismissed):
This change looks good! Thanks for sending it.

Just as a heads up, some other work ongoing at the moment may result in us refactoring this into two configs, rather than using the `options` block to replace `hugo` with `hugo-extended`. Nothing is required from you, but if you check back here in a week it might look different 😆

marshall007:
@imjasonh in hindsight, I think there's a strong case for only including the "extended edition" in wolfi and the image distributions. As noted in the PR description, extended is the recommended install. The homebrew formula, as an example, only includes extended: https://github.com/Homebrew/homebrew-core/blob/603126a323fc3d3fb759e4c558e6b692935832ee/Formula/hugo.rb#L22

The extended binary is only a few MB larger, so I don't see any downside really. The alternative is that we maintain two entirely separate image distributions with all the caveats and differences between the two documented. I'd say 99% of the time folks will run `docker run cgr.dev/chainguard/hugo`, only to realize they have the wrong edition.

jdolitsky (approved):

images/hugo/image.yaml
```diff
@@ -4,6 +4,24 @@ versions:
...
+            - extended
+
+options:
+  extended:
+    contents:
+      packages:
+        add:
+          - hugo-extended
+        remove:
+          - hugo
```
jdolitsky, line 27: It looks like the "remove" section is unfortunately no longer supported. We may need separate apko configs for the extended versions since CI fails with:
```
❕ x86_64    | buildImage failed: installing apk packages: unable to install files for pkg hugo-extended: unable to install file over existing one, different contents: usr/bin/hugo
```

marshall007: @jdolitsky did you see my comment above? I think there is a solid case for only shipping the extended variant in images. Happy to go either way though, just let me know.

jdolitsky: ah i see.. will raise this with the team

PR #688, opened 2023-05-27
Fix ingress e2e test.

dlorenc:
It doesn't completely pass yet, but it gets further!

Fixes:

Related:


ajayk (commented):
The test still needs nginx ingress

ajayk (approved):

PR #691, opened 2023-05-28
Add logic to compute "extra packages" from `{globals,images}.yaml`

mattmoor:


mattmoor:
So this successfully translates `globals.yaml` and `images.yaml` into parts of the matrix:
image

There is exactly one image (php-fpm) that uses something other than "add packages" in options, and I'll file an issue to kill this, so that we can deprecate `options:` in `apko.yaml` 🤞

mattmoor:
Right now this is unused (overridden by the explicit var), but I'll start to take advantage of this in subsequent changes.

monopod/pkg/images/images.go
```diff
@@ -152,7 +164,28 @@ func ListAll(opts ...ListOption) ([]Image, error) {
 			apkoTargetTag := strings.Replace(filepath.Base(apkoConfig), constants.ApkoYamlFileExtension, "", 1)
 			apkoAdditionalTags := strings.Join(variant.Apko.Tags, ",")
 			apkoTargetTagSuffix := ""
+
+			// TODO(mattmoor): Look up each of the build options in "m" and
+			// inline any added packages.
```
mattmoor, line 169: dropping this TODO, which is done

imjasonh (approved):

PR #693, opened 2023-05-28
Start to use the computed "extra packages" from monopod.

mattmoor:
Currently only the terraform workflows are sensitive to this, so I'm adding `crane` (which has a `dev` variant) into the set of things we build with TF.


mattmoor:
Hmm, this doesn't appear to work.

I see the extra packages in `input.json`:
image

... but it doesn't come through the `TF_VAR_` setup:
image

PR #729, opened 2023-05-31
Pass through extra packages.

mattmoor:
With one exception this enables us to stop using build options.


jonjohnsonjr (approved):

imjasonh:
We need to revert this after https://github.com/chainguard-images/images/pull/741

PR #731, opened 2023-05-31
Path to vault binary moved.

amouat:
The path to vault in the underlying package will be moved by https://github.com/wolfi-dev/os/pull/2321

This update corrects the path but *does not* add the fixes for upstream helm chart compatibility (we'll do that separately).

Don't merge this until https://github.com/wolfi-dev/os/pull/2321 lands (will mark draft).

### DIFF
--- a/images/vault/configs/latest.apko.yaml
+++ b/images/vault/configs/latest.apko.yaml
@@ -19,7 +19,7 @@ accounts:
   recursive: true
 
 entrypoint:
-  command: /usr/sbin/vault
+  command: /usr/bin/vault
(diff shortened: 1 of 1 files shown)


imjasonh (approved):
lgtm once the package lands

amouat:
Let's not do this. Instead I propose we move straight to the new version in https://github.com/chainguard-images/images/pull/752

The problem is just running `docker run cgr.dev/chainguard/vault` on this version will fail with `Operation not permitted` and no further information if `--cap-add IPC_LOCK` isn't passed.

Issue #692, opened 2023-05-28
`php-fpm`: stop using `options` for `entrypoint`


mattmoor:
![image](https://github.com/chainguard-images/images/assets/2442466/232a87a9-2c61-4173-8a28-9ddd9cb4eb60)

Currently `php-fpm` is the only image using `options` for anything other than adding packages, which we are moving to another mechanism entirely (trying to kill `options:` in `apko.yaml`).

We should split this "options"-based image to just be another config instead.

cc @imjasonh

mattmoor:
related: https://github.com/chainguard-images/images/pull/691

erikaheidi:
On it!

PR #740, opened 2023-05-31
Fix: moving php-fpm into separate config

erikaheidi:
This PR moves the `php-fpm` image variant into a separate config, addressing #692 .

Fixes: #692 

Related: #691 

## Other Considerations
Current tests don't cover php-fpm.


images/php/configs/fpm.apko.yaml
```diff
@@ -0,0 +1,46 @@
+contents:
```
imjasonh, line 1: Do we want to name this file `latest-fpm.apko.yaml` so it produces the tag `:latest-fpm` (and `-dev`) to match the previous tag?

mattmoor (approved):

PR #751, opened 2023-06-01
Drop the build options blender

mattmoor:
We now plumb through `--package-append` or `TF_VAR_extra_packages` to our builds, so we no longer need to blend it into `apko.yaml`.

After `php-fpm` (thanks @erikaheidi 🙏 ) the only thing build options were used for was to add packages.

We still use `global.yaml` and `images.yaml` to seed the value passed to `--package-append`, but `apko` itself no longed needs this.


jonjohnsonjr (approved):

jdolitsky (approved):
Button click reserved for jon

mattmoor:
Let's wait until we sort out the auth weirdness we are seeing

mattmoor:
things are looking good, so I'm gonna hit the button 🤞

PR #773, opened 2023-06-02
Shift tagging to happen via TF

mattmoor:


mattmoor:
This has the added benefit of testing the tagging presubmit 🤓

mattmoor:
one crane leg: 
image

the other crane leg:
image

tflib/version-tags/main.tf
```diff
@@ -0,0 +1,14 @@
...
+}
+
+variable "package" {
+  type        = string
+  description = "The name of the package from which to extract version tags."
+}
+
+output "tag_list" {
+    value = [
+    for x in var.config.contents.packages : regexall("(((([a-z0-9]+)(?:[.][a-z0-9]+)?)(?:[.][a-z0-9]+)?)(?:[-][a-z0-9]+)?)", trimprefix(x, "${var.package}=")) if startswith(x, "${var.package}=")
```
imjasonh, line 12: Just making sure we still think this usage is compatible with https://github.com/chainguard-dev/terraform-provider-apko/issues/74 which we just need to implement in the provider, right?

imjasonh (approved):

tflib/version-tags/main.tf
mattmoor: Yup!

I put this in its own module in my other PR (pulled in here) to avoid it proliferating.  Using this within the workflow felt like as good a time as any to pull that in.

PR #775, opened 2023-06-02
Use chainguard-dev/actions/matrix-extra-inputs where possible

jdolitsky:


.github/workflows/presubmit-build.yaml
```diff
@@ -40,6 +40,7 @@ jobs:
         run: |
           echo "Error: ${{ matrix.imageName }} image is missing tests."
           exit 1
+      # TODO: Find way to use chainguard-dev/actions/matrix-extra-inputs
```
jdolitsky, line 43: this is a bit more complicated due to alpinelinux.org check below.. have to think a bit how to do this

imjasonh (approved):

PR #758, opened 2023-06-01
[WIP] Switch to dedicated TF files

mattmoor:

PR #781, opened 2023-06-03
Move towards dedicated TF modules per image

mattmoor:
This breaks off a significant chunk of https://github.com/chainguard-images/images/pull/758

This iteration will let us do more interesting things with dedicated test modules (e.g. @imjasonh helm testing) but the dedicated modules are more verbose than I'd like and we don't shed any of the `image.yaml` configuration since these are still invoked individually per variant.

### DIFF
--- a/images/consul/main.tf
+++ b/images/consul/main.tf
@@ -1,0 +1,66 @@
+terraform {
+  required_providers {
+    apko = { source = "chainguard-dev/apko" }
+    oci  = { source = "chainguard-dev/oci" }
+  }
+}
+
+variable "target_repository" {
+  description = "The docker repo into which the image and attestations should be published."
+}
+
+variable "tag_suffix" {
+  description = "The suffix to the tag for this image."
+}
+
+variable "extra_repositories" {
+  type        = list(string)
+  default     = ["https://packages.wolfi.dev/os"]
+  description = "The list of additional repositories to append to the apko configuration."
+}
+
+variable "extra_keyring" {
+  type        = list(string)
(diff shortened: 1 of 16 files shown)


images/consul/main.tf
```diff
@@ -0,0 +1,66 @@
...
+  type        = list(string)
+  default     = ["wolfi-baselayout"]
+  description = "The list of additional packages to append to the apko configuration."
+}
+
+provider "apko" {
+  extra_repositories = var.extra_repositories
+  extra_keyring      = var.extra_keyring
+  default_archs      = ["x86_64", "aarch64"]
+}
```
mattmoor, line 38: Ultimately this is a key piece we will be able to remove with a top-level `main.tf` that configures these ~globally, but I'm opting for the "evil" of duplicating this boilerplate to let us have per-image control sooner than that.

images/ko/configs/latest.apko.yaml
```diff
@@ -1,4 +1,8 @@
 contents:
+  repositories:
```
imjasonh, line 2: I'm confused about the changes to this file, I would have expected this to only have red lines?

mattmoor: Ah, Josh dropped these when he ported `ko` to TF and we haven't been consistent about dropping these everywhere so I brought them back for now.  I'm pretty sure this would pass without this because of the variables, but that was some of my reasoning fwiw.

imjasonh (approved):

PR #786, opened 2023-06-03
use apko publisher module 0.0.4

imjasonh:
We should have dependabot do this, and probably have a script we can run manually to upgrade everything in the repo.

Matt says:
> I think I want a shim for the module in the repo to let us centrally manage that version


mattmoor (approved):
Yeah I’ll add the shim when I get home.  Thanks for fixing this!

PR #798, opened 2023-06-05
Added more in-depth vault test.

amouat:
Added a better vault test. Not sure why the last one was passing :/


amouat:
It doesn't look like the presubmit hook ran the test. It did run the first two. Any suggestions? 

You can see the [run here](https://github.com/chainguard-images/images/actions/runs/5176351988/jobs/9325072421?pr=798)

dlorenc:
> Apparently k8s tests don't run on PRs by default, but I tested on my fork and it looks good.

Can you link the passing run?

imjasonh (approved):

PR #814, opened 2023-06-05
WIP Image: add ingress-nginx-contoller:1.8.0

strongjz:
Fixes:

Related: 

### Pre-review Checklist

- [ ] Image is a smaller size than its common public counterpart, or if not, the reason why should be explained
- [ ] Image should be scanned for CVEs and should return 0 or near-0 results. Exceptions should be explained.
- [ ] The software in the image should be able to be installed and work as expected
  - [ ] The image should start in a cluster
  - [ ] The application should be accessible to the user/cluster/etc. as expected
- [ ] If there is an open source helm chart provided, the image should be able to be used with the chart. If needed, please add a -compat package to close any gaps with the public helm chart.
- [ ] build for x86_64 and aarch64 if possible, document exceptions

#### For new image PRs only

- [ ] Perform manual tests as needed and document your testing results
  - [ ] For builder images (go, python, etc), build a sample app successfully
  - [ ] For services images (rabbit, databases, webservers) test basic functionality, upstream install/getting started, port availability, admin access. Document differences from public image.
  - [ ] Applications intended for use with a specific cloud provider must be tested on that cloud provider.
- [ ] ENTRYPOINT
- [ ] CMD:

- [ ] Add annotations e.g:
```
annotations:
```
#### Documentation Requirements

- [ ] Provide a README file, follow the [README template](readme-template.md) to include variants, tags, and usage examples
- [ ] Document your testing and results in the PR
- [ ] Document any exceptions to the testing and config requirements outlined in this template
- [ ] Document where and how the image deviates from popular alternatives (ports, packages, user, entrypoint, versions, etc). Explain why there's a difference and if the user needs to do anything differently when using the image


henrysachs:
@strongjz can you say a few things about the image that are blocking the merge? I read through your PR and it looks quite promising already. Keep up the great work! 🔥

images/ingress-nginx-controller/configs/latest.apko.yaml
```diff
@@ -0,0 +1,196 @@
...
+    permissions: 0o755
+    recursive: true
+  - path: /sbin/nginx
+    source: /usr/local/nginx/sbin/nginx
+    type: symlink
+    uid: 101
+    gid: 101
+    permissions: 0o755
+  - path: /usr/lib/libmimalloc.so
+    source: /usr/lib/libmimalloc.so.1.7
```
rawlingsj, line 174: symlinks tend to go directly into the apk package, that way we don't have to maintain the soname version here when the mimalloc package update brings in a new version.

images/ingress-nginx-controller/configs/latest.apko.yaml
```diff
@@ -0,0 +1,196 @@
...
+    type: symlink
+    uid: 101
+    gid: 101
+    permissions: 0o755    
+  - path: /usr/local/lib/libmimalloc.so
+    source: /usr/lib/libmimalloc.so.1.7
```
rawlingsj, line 180: symlinks tend to go directly into the apk package, that way we don't have to maintain the soname version here when the mimalloc package update brings in a new version.

rawlingsj (changes requested):

images/ingress-nginx-controller/configs/latest.apko.yaml
```diff
@@ -0,0 +1,196 @@
...
+    - ingress-nginx-controller
+    - dumb-init
+    - geoip-dev
+    - pcre
+    - zlib
+    - libxml2
+    - tzdata
+    - opentracing-cpp
+    - msgpack
+    - mimalloc=1.7.6-r2
```
rawlingsj, line 23: does this need to be pinned to 1.7.6-r2?  I wonder if instead it can be > 1.0 or something so rebuilds will continue to receive minor and patch releases?

images/ingress-nginx-controller/configs/latest.apko.yaml
```diff
@@ -0,0 +1,196 @@
...
+    - lua-resty-dns
+    - lua-resty-core
+    - lua-resty-global-throttle
+    - lua-resty-ipmatcher
+    - lua-resty-memcached
+    - lua-resty-http
+    - lua-resty-lock
+    - lua-resty-redis
+    - lua-resty-string
+    - lua-resty-upload
```
rawlingsj, line 44: maybe these packages can be dropped once [this](https://github.com/wolfi-dev/os/pull/3784/files#diff-433c67391bb6d11409699d4634e68ebd1d199daf95eaf43b0fc745f3a6014927R11-R47) merges as there are runtime deps defined on the main package?

PR #822, opened 2023-06-06
Default to using terraform when unspecified.

mattmoor:
The vast majority of this PR was generated with the following script:
```
#!/bin/bash -e

for x in $(find . -name 'image.yaml'); do

  cp $x $x.bak
  [ "$(grep 'terraform: ' $x)" != "" ] || cat > $x <<EOF
# Opt-out of the new terraform-based release
terraform: false

$(cat $x.bak)
EOF

  rm $x.bak
done
```

The one exception is `monopod/` where we default this field to true.

### DIFF
--- a/images/apko/image.yaml
+++ b/images/apko/image.yaml
@@ -1,3 +1,6 @@
+# Opt-out of the new terraform-based release
+terraform: false
+
 versions:
   - apko:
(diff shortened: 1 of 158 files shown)


jdolitsky (approved):
image

PR #857, opened 2023-06-14
poc terragrunt implementation

joshrwolf:
This is a __POC__ implementation of using [`terragrunt`](https://terragrunt.gruntwork.io/) to remove a lot of the terraform boilerplate being copied around. It isn't a complete implementation, and doesn't update CI accordingly, __DO NOT MERGE__.

Only `ko` and `jdk` are implemented, but this could apply to every image.

The benefits with this approach are:

- a single, shared source of truth for all terraform invocation: `./tflib/image/`
  - each image/tag variant just needs a `terragrunt.hcl` file
  - avoids the need to redefine `main.tf` in every image
- structured configuration overlaying
  - this poc demonstrates layers of overrides according to: `root -> image -> tag`
- easier orchestration without the common trappings of a mono apply
  - build everything with `terragrunt run-all apply`
  - build individual images with `terragrunt apply`

Here's an example of it running:

```bash
➜ terragrunt run-all apply
INFO[0000] The stack at /Users/wolf/dev/gh/chainguard/images will be processed in the following order for command apply:
Group 1
- Module /Users/wolf/dev/gh/chainguard/images/images/jdk/11
- Module /Users/wolf/dev/gh/chainguard/images/images/jdk/latest
- Module /Users/wolf/dev/gh/chainguard/images/images/ko/latest
 
Are you sure you want to run 'terragrunt apply' in each folder of the stack described above? (y/n)y

...

Apply complete! Resources: 11 added, 0 changed, 0 destroyed.

...

Apply complete! Resources: 11 added, 0 changed, 0 destroyed.

...

Apply complete! Resources: 11 added, 0 changed, 0 destroyed.
```

This isn't a full implementation, but it's enough to demonstrate what the end goal could look like. It's also simple enough to act as a "wrapper" around the niceties of the existing terraform implementation, without getting in the way with too much "magic".


root.hcl
```diff
@@ -0,0 +1,52 @@
+locals {
```
joshrwolf, line 1: This file is where the bulk of the work is done.

root.hcl
joshrwolf, line 1: this file is where the bulk of the work is done

images/ko/latest/terragrunt.hcl
joshrwolf, line 1: this represents the bare minimum a tag needs, and essentially "imports" the `root.hcl` default specification

images/ko/image.hcl
joshrwolf, line 1: any "image" level configs/overrides would go here, everything in the `locals` block is merged with the `root.hcl`'s `input` block.

PR #883, opened 2023-06-21
Bulk migrate images without any real complexity to TF

mattmoor:


tuananh:
@mattmoor may I ask why TF is preferred?

mattmoor:
@tuananh for a number of reasons:
1. It moves a bunch of bespoke infrastructure spread across bash, go, yaml and actions, which few people understand into a more mature declarative language (terraform / HCL),
2. We have already slammed into the ceiling of parallelism we can achieve via Github actions (the 256 job matrix limit that prompted sharding a few weeks ago, and the parallelism is still effectively throttled by runner capacity),
3. With the high fanout hitting Github, we see a significant number of failures and rate limits from Github that we cannot pay them more to make go away,
4. By (eventually) consolidating to a single `terraform apply`, we dramatically increase the scope of in-process caching across a highly parallel build.

In our downstream repo we were able to switch from ~40 matrix legs, the longest of which took 12m to a single 4m apply on a 64-core node with a lot of the "overhead" now shared (and much faster due to the larger node).  @jonjohnsonjr tells me he can make it even faster 🎉 

But perhaps most interestingly, without all of the bespoke Github actions infrastructure, we can actually reproduce these builds locally with:
```
terraform init
TF_VAR_target_repository=ghcr.io/mattmoor/tf-root terraform apply
```

I was able to do this for our downstream repo in ~5-6m on my macbook.  You can do this for a single image directory with `-target=module.{foo}`.  I documented a bunch of this in `TERRAFORM.md`, if you want to try it out.

mattmoor:
For grins I just ran this locally:
```
TF_VAR_target_repository=ghcr.io/mattmoor/blah time terraform apply -auto-approve
```

... and it took about 200s (likely with a warm local APK cache)

images/apko/main.tf
```diff
@@ -0,0 +1,37 @@
...
+
+variable "target_repository" {
+  description = "The docker repo into which the image and attestations should be published."
+}
+
+module "latest" {
+  source = "../../tflib/publisher"
+
+  target_repository = var.target_repository
+  config            = file("${path.module}/configs/latest.apko.yaml")
```
luhring, line 15: Just learning... why do we still need this config path defined in `images/apko/image.yaml` if we're defining it for Terraform now?

images/deno/main.tf
luhring, line 1: Not specific to `deno`, and **not a blocking comment**:  I'm noticing a fair amount of repetition among the image HCL code. This isn't a crisis or anything, but it does make me wonder if we have plans to either a) further reduce the repeated code, or b) provide a means of auto-generating the HCL for a new image.

My intent isn't to code golf this to as few lines as possible, but I want to be sensitive to the ongoing maintenance burden at the same time. More repetition can mean more places to remember to update if we make global changes, and more work on a per-image basis as we attempt to scale up our image catalog.

luhring (commented):
Getting excited about this 😁

images/apko/main.tf
mattmoor: @luhring because (for now) we are still using `monopod` for the matrix.  Once we switch to invoking the "mega module" directly (without `-target=module.apko`) we can delete that file entirely.

luhring: Got it! Thanks!

images/deno/main.tf
luhring: Sounds reasonable!

luhring (approved):
🚀

PR #897, opened 2023-06-22
migrate cert-manager* to tf workflow

joshrwolf:


README.md
```diff
@@ -1,7 +1,5 @@
 # Chainguard Images
 
-## :warning::exclamation: On August 16th we will be making changes to how image tags are pulled. Please see [the announcement](https://www.chainguard.dev/unchained/scaling-chainguard-images-with-a-growing-catalog-and-proactive-security-updates) for further details. :exclamation::warning:
```
mattmoor, line 3: you need a new monopod

images/cert-manager-acmesolver/main.tf
```diff
@@ -0,0 +1,76 @@
...
+module "latest-cm-dev" {
+  source = "../../tflib/publisher"
+
+  target_repository = var.target_repository
+
+  config = jsonencode(module.latest.config)
+  extra_packages = concat(module.dev.extra_packages, [
+    "cmctl",
+  ])
+}
```
mattmoor, line 39: I don't think we want both of these.

images/cert-manager-acmesolver/main.tf
```diff
@@ -0,0 +1,76 @@
...
+
+module "test-latest-dev" {
+  source = "./tests"
+  digest = module.latest-dev.image_ref
+}
+
+module "test-latest-cm-dev" {
+  source = "./tests"
+  digest = module.latest-cm-dev.image_ref
+}
```
mattmoor, line 60: same

images/cert-manager-acmesolver/main.tf
```diff
@@ -0,0 +1,76 @@
...
+  depends_on = [
+    module.test-latest,
+    module.test-latest-dev,
+    module.test-latest-cm-dev,
+  ]
+
+  tags = merge(
+    { for t in toset(concat(["latest"], module.version-tags.tag_list)) : t => module.latest.image_ref },
+    { for t in toset(concat(["latest"], module.version-tags.tag_list)) : "${t}-dev" => module.latest-dev.image_ref },
+    { for t in toset(concat(["latest"], module.version-tags.tag_list)) : "${t}-cm-dev" => module.latest-cm-dev.image_ref },
```
mattmoor, line 74: The current semantics puts `cmctl` into the normal `-dev` there isn't another tag

images/cert-manager-acmesolver/tests/main.tf
```diff
@@ -0,0 +1,16 @@
...
+    oci = { source = "chainguard-dev/oci" }
+  }
+}
+
+variable "digest" {
+  description = "The image digest to run tests over."
+}
+
+data "oci_exec_test" "test" {
+  for_each = toset(fileset(path.module, "*"))
```
mattmoor, line 12: That's a fun trick...

mattmoor: I think this will include the `.tf` file.  We should make this `*.sh`

images/cert-manager-cainjector/image.yaml
```diff
@@ -1,21 +1,5 @@
-# Opt-out of the new terraform-based release
-terraform: false
-
-status: experimental
 versions:
   - apko:
       config: configs/latest.apko.yaml
       extractTagsFrom:
         package: cert-manager-cainjector
```
mattmoor, line 5: I've been dropping this as well

mattmoor (commented):
Most of my comments apply to all of them, but otherwise this looks really close 😁

images/cert-manager-acmesolver/tests/main.tf
joshrwolf: good catch, I was planning on lifting it out into a module, but will fix it here for now

README.md
```diff
@@ -17,10 +17,7 @@
...
-| [cert-manager-webhook](./images/cert-manager-webhook) | `cgr.dev/chainguard/cert-manager-webhook` | stable | [![](https://storage.googleapis.com/chainguard-images-build-outputs/badges/cert-manager-webhook.build.status.latest-dev.svg)](https://registry-ui.chainguard.app/?image=cgr.dev/chainguard/cert-manager-webhook:latest-dev)<br/>[![](https://storage.googleapis.com/chainguard-images-build-outputs/badges/cert-manager-webhook.build.status.latest.svg)](https://registry-ui.chainguard.app/?image=cgr.dev/chainguard/cert-manager-webhook:latest) |
+| [cert-manager](./images/cert-manager) | `cgr.dev/chainguard/cert-manager` | stable | [![](https://storage.googleapis.com/chainguard-images-build-outputs/badges/cert-manager.build.status.latest.svg)](https://registry-ui.chainguard.app/?image=cgr.dev/chainguard/cert-manager:latest) |
```
joshrwolf, line 20: this is ~wrong, but I can update `monopod` to do this properly if we're ok with this approach

imjasonh (dismissed):
I love this. I definitely want to do this elsewhere, thanks for this example.

images/cert-manager/main.tf
```diff
@@ -0,0 +1,75 @@
...
+  source = "./tests"
+
+  digests = { for k, v in module.latest : k => v.image_ref }
+}
+
+module "test-latest-dev" {
+  source = "./tests"
+
+  digests   = { for k, v in module.latest : k => v.image_ref }
+  suffix    = "-dev"
```
mattmoor, line 57: I’d use a book instead of a suffix if we can

mattmoor: Bool

.github/workflows/presubmit-build.yaml
```diff
@@ -76,6 +76,10 @@ jobs:
         if: ${{ ! matrix.useTerraform }}
         with: ${{ fromJSON(steps.augmented-inputs.outputs.matrix-json) }}
 
+      - name: Collect diagnostics and upload
```
mattmoor, line 79: This is super useful thanks for adding this 🙏

images/cert-manager/image.yaml
```diff
@@ -0,0 +1,3 @@
+versions:
+  - apko:
+      config: configs/latest.apko.yaml
```
mattmoor, line 3: This is going to mess with the post-submit "policy verification" logic.

IMO we should file an issue to bring back some form of policy verification, and then just rip it out for now.

The place we're (currently) doing it is ~wrong anyways because it's after we've tagged stuff, so it won't block stuff like catalog syncing.

images/cert-manager/tests/main.tf
```diff
@@ -0,0 +1,78 @@
...
+  description = "Suffix to append to the helm release name."
+  default     = ""
+}
+
+variable "skip_crds" {
+  default = false
+}
+
+resource "helm_release" "cert-manager" {
+  name             = "cert-manager${var.suffix}"
```
mattmoor, line 28: Can we make this generate a random suffix instead (see what Jason did in his PR)?

images/cert-manager/main.tf
mattmoor: Actually, do you even need this if you just randomize the suffix?

images/cert-manager/image.yaml
joshrwolf: https://github.com/chainguard-images/images/issues/941

mattmoor (approved):

PR #901, opened 2023-06-22
Migrate the Ruby images to TF

mattmoor:


images/ruby/3.0.tf
```diff
@@ -0,0 +1,37 @@
...
+
+  target_repository = var.target_repository
+  # Make the dev variant an explicit extension of the
+  # locked original.
+  config = jsonencode(module.three-zero.config)
+  extra_packages = concat(module.dev.extra_packages, [
+    "build-base",
+    "ruby-3.0-dev",
+    "ruby3.0-bundler",
+  ])
```
imjasonh, line 19: I made this a local `node_dev_packages`, that might be easier than copying this.

imjasonh (approved):

images/ruby/3.0.tf
mattmoor: The problem is the version numbers :-/

PR #905, opened 2023-06-22
use hugo-extended package in image

marshall007:
Related: https://github.com/chainguard-images/images/pull/682

### Pre-review Checklist

- [ ] Image is a smaller size than its common public counterpart, or if not, the reason why should be explained
- [ ] Image should be scanned for CVEs and should return 0 or near-0 results. Exceptions should be explained.
- [ ] The software in the image should be able to be installed and work as expected
  - [ ] The image should start in a cluster
  - [ ] The application should be accessible to the user/cluster/etc. as expected
- [ ] If there is an open source helm chart provided, the image should be able to be used with the chart. If needed, please add a -compat package to close any gaps with the public helm chart.
- [ ] build for x86_64 and aarch64 if possible, document exceptions

### DIFF
--- a/images/hugo/configs/latest.apko.yaml
+++ b/images/hugo/configs/latest.apko.yaml
@@ -1,6 +1,6 @@
 contents:
   packages:
-    - hugo
+    - hugo-extended
 
 entrypoint:
   command: /usr/bin/hugo
--- a/images/hugo/main.tf
+++ b/images/hugo/main.tf
@@ -29,7 +29,7 @@ module "latest-dev" {
 
 module "version-tags" {
   source  = "../../tflib/version-tags"
-  package = "hugo"
+  package = "hugo-extended"
   config  = module.latest.config
 }
 


images/hugo/configs/latest.apko.yaml
```diff
@@ -1,6 +1,6 @@
 contents:
   packages:
-    - hugo
+    - hugo-extended
```
marshall007, line 3: We should consider adding `go` and `git` at minimum but possibly `nodejs` and the `autoprefixer` and `postcss` npm packages as well.

Example: https://github.com/klakegg/docker-hugo/blob/b5dfd68114f927fb9cee2e22215a485b6c51b26d/src/docker/alpine-ext/ext-alpine.df#L34

`go` and `git` are required for `hugo mod` subcommands and the `npm` packages are required for the PostCSS features of the extended Hugo build to work.

marshall007:
@jdolitsky FYI I'm finally getting back around to this from previous attempt in https://github.com/chainguard-images/images/pull/682.

marshall007:
TODO I think running a production build of the docsy example site would make for a more robust test: https://github.com/google/docsy-example

For context, Docsy is the theme that the majority of CNCF projects use for their doc sites.

images/hugo/configs/latest.apko.yaml
mattmoor: One thing we might consider here is to have the `-dev` variant include these on top of the standard corpus of "dev" packages.  We extend things like this in a number of images (e.g. `python` with `pip`, `nodejs` with `yarn`(?), `cert-manager` with `cmctl`, etc).

marshall007: @mattmoor in this particular case I wouldn't really consider most of these "dev" dependencies. See [installation prerequisite docs](https://gohugo.io/installation/linux/#prerequisites):

- `go` + `git`: Use the [Hugo Modules](https://gohugo.io/hugo-modules/) feature (requires `git` for same reasons `go mod` does)
- `git`: Install a theme as a Git submodule (predates Hugo Modules)
- `git`: Access [commit information](https://gohugo.io/variables/git) from repo (`.GitInfo` page variables)

As of [Hugo `v0.114.0`](https://github.com/gohugoio/hugo/releases/tag/v0.114.0) (released 3 days ago) installing `dart-sass` separately is preferred and the embedded `libsass` is deprecated. See [sass compilation docs](https://gohugo.io/hugo-pipes/transpile-sass-to-css/#dart-sass).

Finally, NodeJS (and related npm packages) are required for CSS post-processing features. See [PostCSS docs](https://gohugo.io/hugo-pipes/postcss/#setup). This part is pretty annoying because most Hugo sites and popular themes need these tools to build in `--production` mode.

If you poke around you will find a ton of (even high profile) projects that blindly install the latest versions of these packages globally at build time:

- https://github.com/Azure/azure-service-operator/blob/10e6263dd818b177b4fab2717b8c27f0adcc0ee2/.devcontainer/install-dependencies.sh#L220-L225
- https://github.com/GoogleContainerTools/skaffold/blob/ce682a532dd4bcfc1efcb1320975cd2e86740ddd/deploy/webhook/Dockerfile#L40-L45
- https://github.com/kairos-io/kairos/blob/f883e8ec7ed0d8021c0d3b7b0e862502e56cb5ad/Earthfile#L956
- https://github.com/cortexproject/cortex/blob/master/build-image/Dockerfile#L11

From an SSCS perspective I think there is a legitimate problem to be solved here. In addition to the security concern, we've also found that running `npm i -g postcss postcss-cli autoprefixer` brittle in practice as these packages have a tendency to break on latest and across Node versions.

marshall007: @mattmoor to keep this moving along I'm going to add `go` and `git` as hard dependencies (all variants) and include `nodejs` + `npm` in the `-dev` variant. Does that make sense for now?

marshall007:
Closing in favor of https://github.com/chainguard-images/images/pull/1130

PR #906, opened 2023-06-22
migrate node to TF

imjasonh:


images/node/18.tf
```diff
@@ -0,0 +1,32 @@
+module "eighteen" {
+  source = "../../tflib/publisher"
+
+  target_repository = var.target_repository
+  config            = file("${path.module}/configs/20.apko.yaml")
```
mattmoor, line 5: 18?

images/node/image.yaml
```diff
@@ -3,38 +3,8 @@ terraform: false
...
       config: configs/19.apko.yaml
-      extractTagsFrom:
-        package: nodejs-19
-      subvariants:
-        - suffix: -dev
-          options:
-            - dev
-            - node-dev
   - apko:
       config: configs/20.apko.yaml
```
mattmoor, line 10: We should kill these because they all do the same thing, and it'll spawn 3 competing legs

mattmoor (dismissed):

mattmoor (approved):

images/node/example/Dockerfile
```diff
@@ -1,7 +1,10 @@
 # syntax=docker/dockerfile:1
 # Based on https://docs.docker.com/language/nodejs/build-images/
 
-FROM cgr.dev/chainguard/node
+ARG BASE=cgr.dev/chainguard/node
+
+FROM ${BASE}
+
```
imjasonh, line 7: We had not been effectively testing the just-built image, and instead only tested the already-tagged-`:latest` image on every change. This is fixed now!

PR #917, opened 2023-06-23
remove extraneous env var checks

imjasonh:
We already `set -o nounset` which fails when the variable is used but is not set. Removing this boilerplate before it can be copypasted again.

### DIFF
--- a/images/metacontroller/tests/02-helm.sh
+++ b/images/metacontroller/tests/02-helm.sh
@@ -196,7 +196,7 @@ EOF
 
 kubectl -n hello apply -f hello.yaml
 
-# Sleep  theh ello controller kicks in
+# Sleep the hello controller kicks in
 sleep 30
 
 kubectl get pods chainguard --output=jsonpath='{.status.phase}' -n hello | grep "Succeeded"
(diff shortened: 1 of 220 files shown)


images/metacontroller/tests/02-helm.sh
```diff
@@ -198,7 +196,7 @@ EOF
 
 kubectl -n hello apply -f hello.yaml
 
-# Sleep  theh ello controller kicks in 
+# Sleep  theh ello controller kicks in
```
mattmoor, line 199:
```suggestion
# Sleep the hello controller kicks in
```

mattmoor (approved):

images/metacontroller/tests/02-helm.sh
mattmoor: I think merging this will clear my LGTM, and this PR is merge conflict catnip, so let's merge this as-is and follow up.

imjasonh: Bah I committed before reading your comment. If you're around to approve again cool, otherwise I'll deal with the consequences tomorrow.

ajayk (approved):

PR #920, opened 2023-06-23
Port fluentd to TF

mattmoor:


images/fluentd/tests/main.tf
```diff
@@ -0,0 +1,14 @@
...
+  }
+}
+
+variable "digest" {
+  description = "The image digest to run tests over."
+}
+
+data "oci_exec_test" "test" {
+  digest = var.digest
+  script = "${path.module}/01-test.sh"
```
imjasonh, line 13: Let's use `working_dir` so the test doesn't have to know about the structure of the repo above this directory.

mattmoor: The files it is accessing are above this directory (`examples` is generally a sibling).  I think my preference would be to move `examples` under the test directory, but this confuses `monopod` right now.  I was hoping we could do some broader cleanup once we're off of `monopod matrix` 🤞 

I can change it, it just felt a little gross to specify a `working_dir` and then use `../examples` 🤢

mattmoor: tl;dr I'd love to work towards relatively self-contained modules 🤞

mattmoor: https://github.com/chainguard-images/images/issues/925

imjasonh (approved):

PR #922, opened 2023-06-23
Add bash as default cmd.

amouat:
Update to make `docker run -it cgr.dev/chainguard/bash` work as expected.


cpanato (approved):
i think makes sense

amouat:
Thanks. I'll see if I can add some tests per @imjasonh's comments.

amouat:
Added a really simple test just to make sure it doesn't error.

amouat:
I converted to draft while we figure out how to handle https://github.com/chainguard-images/images/issues/923

PR #927, opened 2023-06-23
withdraw some more php tags

imjasonh:
These were also erroneously published, and should be withdrawn.

Some of these, like `musl` and `glibc` tags, are artifacts of how the PHP image used to be built on Alpine, and was migrated to Wolfi a long time ago.


mattmoor (dismissed):

erikaheidi (approved):

mattmoor (approved):
🙏

PR #928, opened 2023-06-23
port argocd{-repo-server} to TF (w/ helm test!)

imjasonh:
This ports argocd and argocd-repo-server to TF, with `helm_release` resources that run the newly built image on a cluster using the upstream Helm chart, and waits for the job to become `Ready`.

We run `argocd{-repo-server}{-dev}` images in different namespaces so they don't conflict.

Tested with:

```
terraform apply -target=module.argocd -target=module.argocd-repo-server -var=target_repository=ttl.sh/jason
```

```
$ crane ls ttl.sh/jason/argocd
2-dev
2.7-dev
2.7.6-dev
2.7.6-r1-dev
2.7.6-r1
2.7.6
2.7
2
latest-dev
latest
```

(same for `argocd-repo-server`)


images/argocd-repo-server/tests/main.tf
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+variable "digest" {
+  description = "The image digest to run tests over."
+}
+
+data "oci_exec_test" "run" {
+  digest = var.digest
+  script = "docker run --rm $${IMAGE_NAME} --help"
```
mattmoor, line 14: I was wondering when someone would do this :)

images/argocd-repo-server/tests/main.tf
```diff
@@ -0,0 +1,46 @@
...
+}
+
+resource "random_string" "random" {
+  length  = 6
+  upper   = false
+  special = false
+}
+
+resource "helm_release" "argocd-repo-server" {
```
mattmoor, line 23: I want to confirm this waits for a meaningful installation signal?

mattmoor (dismissed):

images/argocd-repo-server/tests/main.tf
joshrwolf: 👍 the meaningfulness of the signal will depend on the chart, but at a bare minimum the provider defaults to waiting for all resources to be `Ready`

images/argocd-repo-server/tests/main.tf
```diff
@@ -0,0 +1,46 @@
...
+  depends_on = [data.oci_exec_test.run]
+
+  name = "argocd-repo-server-${random_string.random.result}"
+
+  repository = "https://argoproj.github.io/argo-helm"
+  chart      = "argo"
+
+  namespace        = "argocd-repo-server-${random_string.random.result}"
+  create_namespace = true
+  skip_crds        = true
```
joshrwolf, line 33: nit: leave a note here so this argocd specific incantation doesn't get copy/pasta'd everywhere

images/argocd/tests/main.tf
```diff
@@ -0,0 +1,45 @@
...
+
+resource "helm_release" "argocd" {
+  name = "argocd-${random_string.random.result}"
+
+  repository = "https://argoproj.github.io/argo-helm"
+  chart      = "argo"
+
+  namespace        = "argocd-${random_string.random.result}"
+  create_namespace = true
+  skip_crds        = true
```
joshrwolf, line 32:
```suggestion
  # This is argocd specific. 
  # The argocd helm chart installs CRDs from the `templates/` directory, and the default value (`false`) conflicts with the providers assumption of them being in `crds/`
  skip_crds        = true
```

images/argocd-repo-server/tests/main.tf
imjasonh: There's a `wait` parameter that's true by default, which waits for the jobs to become ready.

joshrwolf (approved):

images/argocd-repo-server/tests/main.tf
imjasonh: There's also `wait_for_jobs` (default false) that

> will wait until all Jobs have been completed before marking the release as successful. It will wait for as long as timeout. Defaults to false.

images/argocd-repo-server/tests/main.tf
imjasonh: Part of my desire to remove boilerplate was to make it easier to tell when this was possible. I'll probably make a pass at more of them in a bit.

PR #931, opened 2023-06-23
migrate kubectl to tf workflow

joshrwolf:
This depends on #928 landing since it re-uses the shared kind cluster.


images/kubectl/configs/latest.apko.yaml
```diff
@@ -1,12 +1,6 @@
 contents:
-  keyring:
-    - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
-  repositories:
-    - https://packages.wolfi.dev/os
   packages:
-    - ca-certificates-bundle
     - kubectl
```
mattmoor, line 3: doesn't get old :)

mattmoor (approved):

PR #935, opened 2023-06-25
Drop `apkoBuildOptions`

mattmoor:
We stopped using these a while ago when we adopted `--packages-append` and removed the `yq` logic that mixed-in the options to the `apko` configuration files, but we never dropped this piece, which should now be superfluous.

This cleans up the remnants of `apkoBuildOptions` in images.


mattmoor:
Overriding, since `fluentd` is broken at HEAD

tstromberg (commented):
LGTM

PR #936, opened 2023-06-25
Fix the "extra" overrides.

mattmoor:
When we switched to the mega module, we lost our sensitivity to these things: https://github.com/chainguard-images/images/blob/b9efef911e8e00ec6e20749fa3c77f148a99e8f2/.github/actions/release-image-terraform/action.yml#L69-L79

This brings it back!

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -14,10 +14,25 @@ variable "target_repository" {
   description = "The docker repo into which the image and attestations should be published."
 }
 
+variable "extra_repositories" {
+  type    = list(string)
+  default = []
+}
+
+variable "extra_keyring" {
+  type    = list(string)
+  default = []
+}
+
+variable "extra_packages" {
+  type    = list(string)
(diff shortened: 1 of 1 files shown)


main.tf
```diff
@@ -14,10 +14,22 @@ variable "target_repository" {
   description = "The docker repo into which the image and attestations should be published."
 }
 
+variable "extra_repositories" {
+  default = []
+}
+
+variable "extra_keyring" {
```
jonjohnsonjr, line 21: It kills me that this one is singular and the others are plural.

mattmoor: I am hoping we can kill it all before too long, but you have to make it stupid fast first :)

mattmoor:
Hmm, I'm not sure why these list arguments are coming through as strings 🤔

imjasonh (approved):

PR #942, opened 2023-06-26
migrate dex to TF w/ helm test

imjasonh:


images/dex/tests/runs.sh
```diff
@@ -5,13 +5,10 @@ set -o errexit -o nounset -o errtrace -o pipefail -x
 PORT=5556
 
 URL="http://127.0.0.1:${PORT}"
```
mattmoor, line 7: It looks like this runs on the host below
```suggestion
URL="http://127.0.0.1:${FREE_PORT}"
```

mattmoor (dismissed):

images/dex/tests/runs.sh
```diff
@@ -4,14 +4,11 @@ set -o errexit -o nounset -o errtrace -o pipefail -x
 
 PORT=5556
 
-URL="http://127.0.0.1:${PORT}"
-docker run --rm -d -p $PORT:$PORT "${IMAGE_NAME}" dex serve /etc/dex/config.docker.yaml
+URL="http://127.0.0.1:${FREE_PORT}"
```
mattmoor, line 7: We have an asymmetry.  We need to `curl` with `FREE_PORT`, but the issuer output on line `29` should grep the URL with `PORT`.

The current script can use a single `URL` variable because the port are symmetric, but we're changing that 😞

mattmoor (approved):

PR #946, opened 2023-06-26
migrate flux to consolidated terraform

joshrwolf:
depends on https://github.com/wolfi-dev/os/pull/3126

consolidates all `flux*` images into a single module, with the appropriate helm test


images/flux/tests/main.tf
```diff
@@ -1,20 +1,58 @@
...
+  repository       = "https://fluxcd-community.github.io/helm-charts"
+  chart            = "flux2"
+  create_namespace = true
+
+  values = [
+    <<EOF
+cli:
+  image: ${data.oci_string.ref["cli"].registry_repo}
+  tag: ${data.oci_string.ref["cli"].pseudo_tag}
+helmController:
```
mattmoor, line 38: Is there a reason we do it this way vs. the other way?  If so, a comment would be helpful to explain why.

generally for structured stuff like this I like to use `jsonencode({...})` since json is valid yaml and it let's us shed the `${}` wrapping, and gets some measure of syntax prevalidation.

mattmoor (approved):
Just a question/nit otherwise this looks amazing!

images/flux/tests/main.tf
joshrwolf: no good reason, old habbits die hard 😄 

although you got me experimenting with `jsonencode` and I like that better in this case.

I'll land this as is for now, and follow up with a PR shortly after to convert these over alongside the shift to `data.oci_string`

joshrwolf: nevermind, ran into a merge conflict so changing it here 🎉

imjasonh (approved):

PR #950, opened 2023-06-26
migrate kyverno and friends to consolidated tf

joshrwolf:
depends on final implementation of: https://github.com/chainguard-dev/terraform-provider-oci/pull/47


joshrwolf:
closed in favor of #972 , but might clean this up with the `data.oci_name` thing later

PR #954, opened 2023-06-26
Add a test that we can reproduce the images we build.

mattmoor:


imjasonh (approved):

mattmoor:
I'm bypassing the `fluentd` presubmit because it is failing at HEAD, so I can babysit this postsubmit.

PR #951, opened 2023-06-26
migrate busybox to terraform

joshrwolf:
the one interesting thing here is the use of an aliased `provider` to allow using an alpine specific repo, which the `musl` based `busybox` requires.

PR #960, opened 2023-06-27
migrate git to terraform workflow

joshrwolf:
in line with how #951 does the provider alias

the hardest part with this is the tagging structure, which is heavily commented to help readability.

worth noting that this __adds__ additional `root` variants for the `root` variant with the following new tags:

```bash
latest-root-dev
root-$v-dev
```

this is done to simplify the conditional logic (it's more complicated to make a conditional to avoid `-dev` for the `root` variant than to just build it)


images/git/main.tf
```diff
@@ -0,0 +1,81 @@
...
+    # Tagging format for alpine/musl variant:
+    # - latest
+    # - latest-dev
+    # - latest-root
+    # - latest-root-dev
+    # - $v
+    # - $v-dev
+    # - root-$v
+    # - root-$v-dev
+    (each.key == "nonroot" ?
```
joshrwolf, line 41: I'm not super happy with this approach, since if we ever extend this to >2 this ternary will fall apart. but this was the "cleanest" way I could come up with

images/git/alpine.tf
```diff
@@ -0,0 +1,47 @@
...
+  providers = {
+    apko = apko.alpine
+  }
+  for_each = local.accounts
+  source   = "../../tflib/publisher"
+
+  target_repository = var.target_repository
+  config            = file("${path.module}/configs/latest.alpine.${each.key}.apko.yaml")
+  # Override the module's default wolfi packages that conflict with alpine
+  extra_packages = module.dev.extra_packages
```
joshrwolf, line 24: the current list of wolfi dev variants have suitable alpine replacements, but this could come back to haunt us

imjasonh: Just to confirm, if we added something to dev.extra_packages that was only present in Wolfi (or only in Alpine), the mega-apply would fail relatively loudly, before testing or especially tagging anything right? I think that's how the haunting would manifest now, but at least we wouldn't tag things inconsistently between Alpine/Wolfi and have to unwind that (right?)

I don't have any good ideas to ensure we keep that guarantee, it's just something to be aware of. I don't foresee many huge changes to dev.extra_packages, and if we needed to I think we could come up with something.

images/git/image.yaml
```diff
@@ -1,36 +1,10 @@
-# Opt-out of the new terraform-based release
-terraform: false
-
 versions:
```
imjasonh, line 1: I think we're supposed to end up with one item in this list. Is there a reason this has to have an item for each yaml file?

images/git/main.tf
imjasonh: We've talked about moving accounts and run-as into TF config so it can be shared more easily, this might be worth doing sooner if this makes this simpler.

imjasonh: These comments help a lot, thanks for adding them!

images/git/alpine.tf
joshrwolf: correct, this would surface as an apply-time error for one of the following reasons:

1. a dev wolfi package is defined that doesn't exist 1:1 in alpine
2. a dev wolfi package contains conflicts with an identically named package in alpine

either way, I think we'd catch it before shipping anything, so I think this is ok to leave for now

images/git/main.tf
```diff
@@ -0,0 +1,81 @@
...
+}
+
+variable "target_repository" {
+  description = "The docker repo into which the image and attestations should be published."
+}
+
+module "dev" { source = "../../tflib/dev-subvariant" }
+
+module "tagger" {
+  for_each = local.accounts
```
mattmoor, line 21: I always worry about multiple `tagger` invocations since `tagger` exists to avoid tag collisions.  I wonder if it's worth trying to turn it into one giant map instead?  We already wait on all tests in each variant.

joshrwolf: I flattened it into a single module and am much happier with the outcome! I think that's about as readable as it'll get

main.tf
```diff
@@ -33,9 +33,10 @@ provider "apko" {
 provider "apko" {
   alias = "alpine"
 
-  default_archs      = [] # this defaults to "all"
   extra_repositories = ["https://dl-cdn.alpinelinux.org/alpine/edge/main"]
-  extra_packages     = ["ca-certificates-bundle"]
+  # These packages match chainguard-images/static
+  extra_packages = ["alpine-baselayout-data", "alpine-release", "ca-certificates-bundle"]
```
joshrwolf, line 38: I think these make sense at the provider level, to mimic what we do with wolfi (`wolfi-baselayout` and `ca-certificates-bundle`)

mattmoor: Me too, can you scrub them from the other alpine `.apko.yaml` files?

images/git/main.tf
```diff
@@ -0,0 +1,70 @@
...
+    # - latest-dev
+    # - latest-root
+    # - latest-root-dev
+    # - $v
+    # - $v-dev
+    # - root-$v
+    # - root-$v-dev
+    # Nonroot alpine variant
+    { for t in toset(concat(["latest"], module.version-tags-alpine["nonroot"].tag_list)) : t => module.latest-alpine["nonroot"].image_ref },
+    { for t in toset(concat(["latest"], module.version-tags-alpine["nonroot"].tag_list)) : "${t}-dev" => module.latest-alpine-dev["nonroot"].image_ref },
```
mattmoor, line 42: I think my only micro-nit would be that, given that we use `["..."]` as a subscript, and this is the only version that does the `concat`, I wonder if it would be slightly clearer to do `latest` explicitly on its own rows.

That said, I agree this is about as clear as we're going to get 🙏

mattmoor (dismissed):

images/git/main.tf
joshrwolf: no nit is too small!

mattmoor (approved):

PR #963, opened 2023-06-27
migrate trust-manager to TF

imjasonh:


joshrwolf:
I worry about the coupling here, if jetstack decides they want another component then our ternaries fall apart.

does it make more sense to have `trust-manager` be a separate module within `cert-manager`? maybe something like `images/cert-manager/trust-manager` thats imported by `cert-manager/main.tf`?

there'd be more hcl, but I think it might be more sustainable

mattmoor:
@imjasonh This will also rename the image to `certmanager-trust-manager` which I think is a misnomer.

Let's try to descope this to just terrafying `trust-manager` and we can consolidate as a follow-up if it makes sense.

PR #964, opened 2023-06-27
don't use FREE_PORT to get unique container name

imjasonh:
When we use `${FREE_PORT}` to get a ~random container name, without also running something on that port to "claim" it, we end up getting name conflicts and flaky tests.

Instead of relying on `${FREE_PORT}` when we're not going to ~immediately run on that port, just use `${RANDOM}`.


mattmoor (approved):
Can you send the same to the downstream repo? 🙏

imjasonh:
> Can you send the same to the downstream repo? 🙏

https://github.com/chainguard-images/images-private/pull/465

PR #965, opened 2023-06-27
Update remaining container name references to use hoisted variable

mattmoor:


imjasonh (approved):

mattmoor:
Everything is passing except `fluentd` which is failing at HEAD, so I am going to bypass the presubmit check.

PR #977, opened 2023-06-28
use apko_tags, slay the regexp monster

imjasonh:
This produces the same list of tags as we got from the regexp monster previously, but comes with tests in the tf-apko repo and is written in Go.


imjasonh:
The `version-tags` module is now just a wrapper for `apko_tags`. If we want to we can rip it out and inline `apko_tags`, but I don't know if it's worth it right now.

mattmoor (approved):

PR #981, opened 2023-06-28
Add proxysql

Dentrax:
> **Note**
> **Before:** `380MB, 161 packages, 187 vulnerabilities`
> **After:** `29.6MB, 36 packages, 0-CVE`

> **Warning**
> Image does not respects `SIGINT`.

Dockerfile: 
* https://explore.ggcr.dev/?blob=proxysql/proxysql@sha256:5245ece4cd127f9b807a9374fbc5ab637e6c70e1cac28fb6f580c110e864c27d&mt=application%2Fvnd.docker.container.image.v1%2Bjson&size=3506&manifest=proxysql/proxysql@sha256:46dc8ffdbfcaa1657597190cb9fe3a555444d42a00c367103954aeb143a4afaa
- https://github.com/sysown/proxysql/blob/v2.x/docker/images/proxysql/proxysql/compile_and_start_proxysql.sh

Fixes:

Related: 

### Pre-review Checklist

- [X] Image is a smaller size than its common public counterpart, or if not, the reason why should be explained
- [X] Image should be scanned for CVEs and should return 0 or near-0 results. Exceptions should be explained.
- [X] The software in the image should be able to be installed and work as expected
  - [X] The image should start in a cluster
  - [X] The application should be accessible to the user/cluster/etc. as expected
- [X] If there is an open source helm chart provided, the image should be able to be used with the chart. If needed, please add a -compat package to close any gaps with the public helm chart.
- [X] build for x86_64 and aarch64 if possible, document exceptions

#### For new image PRs only

- [X] Include tests, sufficient enough that you would trust this image running in production.
- [X] Perform manual tests as needed and document your testing results
  - [ ] For builder images (go, python, etc), build a sample app successfully
  - [X] For services images (rabbit, databases, webservers) test basic functionality, upstream install/getting started, port availability, admin access. Document differences from public image.
  - [ ] Applications intended for use with a specific cloud provider must be tested on that cloud provider.

- [X] The version included is the latest GA version of the software
- [X] The latest tag points to the newest stable version
- [X] The image runs as `nonroot` and GID/UID are set to 65532 or upstream default
  - [X] Alternatively the username and GID/UID may be a commonly used one from the ecosystem e.g: postgres
- [X] The image contains the ca-certs bundle if needed
- [X] ENTRYPOINT
  - [X] For applications/servers/utilities call main program with no arguments e.g. [redis-server]
- [ ] CMD:

- [X] Add annotations e.g:
```
annotations:
```
- [X] Check if environment variables are needed e.g. to set data locations
- [X] Ensure the image responds to SIGTERM
  - [X] `docker kill $(docker run -d --rm cgr.dev/chainguard/nginx)`
- [X] Error logs write to stderr and normal logs to stdout. DO NOT write to file.
#### Documentation Requirements

- [X] Provide a README file, follow the [README template](readme-template.md) to include variants, tags, and usage examples
- [ ] Document your testing and results in the PR
- [ ] Document any exceptions to the testing and config requirements outlined in this template
- [ ] Document where and how the image deviates from popular alternatives (ports, packages, user, entrypoint, versions, etc). Explain why there's a difference and if the user needs to do anything differently when using the image


images/proxysql/tests/main.tf
```diff
@@ -0,0 +1,14 @@
...
+}
+
+variable "digest" {
+  description = "The image digest to run tests over."
+}
+
+data "oci_exec_test" "version" {
+  digest = var.digest
+  script = "${path.module}/01-version.sh"
+}
```
imjasonh, line 14: I think we can migrate the helm test in this PR. There are some examples in the repo now, for example https://github.com/chainguard-images/images/blob/main/images/stakater-reloader/tests/main.tf#L21

images/proxysql/tests/01-version.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -o errexit -o nounset -o errtrace -o pipefail -x
+
+docker run --rm "${IMAGE_NAME}" --version
```
imjasonh, line 5: We can just put this directly in the `oci_exec` resource now, since this test doesn't do much besides just run the image.

imjasonh (commented):
I think at the end of this PR we can get rid of both of the shell scripts, and the only thing in `tests` is `main.tf`.

images/proxysql/tests/main.tf
Dentrax: It does make sense! I didn't notice this. So how can we ensure pods are up and running? We used to use `kubectl wait ...`, in this case, maybe I can use [wait](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#wait). 🤔

Dentrax: Migrated to TF! Could you please check? @imjasonh

imjasonh (changes requested):
Needs an update to the root main.tf

Dentrax:
Rebased.

dlorenc:
Do you need help here?

developer-guy:
We will make CI passes 🚀

images/proxysql/tests/main.tf
```diff
@@ -0,0 +1,37 @@
...
+    helm   = { source = "hashicorp/helm" }
+  }
+}
+
+variable "digest" {
+  description = "The image digest to run tests over."
+}
+
+
+# we commented out the Helm tests since the Helm Chart uses the tag in labels, so, once we use pseudo tag, it gives an error like the following:
```
imjasonh, line 14: Add this link https://github.com/dysnix/charts/blob/main/dysnix/proxysql/templates/_helpers.tpl#L41

And open an issue to track finding a fix for this.

jdolitsky: opened https://github.com/chainguard-images/images/issues/1065

jdolitsky (approved):

PR #987, opened 2023-06-28
ci: terraform: print diff to stdout if check fails

Dentrax:
Currently, in case `terraform` format check fails, it does not print the _diff_ to output. It'd be nice to display diffs of formatting changes in CI (in case if fail).

Actual output:
```
Run terraform fmt -check -recursive
  terraform fmt -check -recursive
  shell: /usr/bin/bash -e {0}
images/proxysql/tests/main.tf
Error: Process completed with exit code 3.
```


imjasonh (approved):

imjasonh:
CI failures are unrelated.

PR #996, opened 2023-06-29
WIP: add killability test

imjasonh:
Today we have a PR checklist item that says:

> - [ ] Ensure the image responds to SIGTERM
>   - [ ] `docker kill $(docker run -d --rm cgr.dev/chainguard/nginx)`

This seemed easy enough to add as a build test alongside check-reproducibility.sh, so I did that.

...And a bunch of images don't pass (`nginx` does at least).

Opening this to better understand this issue, and see if we can move this from a human checklist item into a computer presubmit check.

### DIFF
--- a/tflib/publisher/main.tf
+++ b/tflib/publisher/main.tf
@@ -43,6 +43,13 @@ data "oci_exec_test" "check-reproducibility" {
   timeout_seconds = 600
 }
 
+data "oci_exec_test" "check-killability" {
+  digest = module.this.image_ref
+  script = "docker kill $(docker run -d --rm $${IMAGE_NAME})"
+
+  timeout_seconds = 30
+}
+
 output "image_ref" {
   value = data.oci_exec_test.check-reproducibility.tested_ref
 }


tflib/publisher/main.tf
```diff
@@ -43,6 +43,13 @@ data "oci_exec_test" "check-reproducibility" {
   timeout_seconds = 600
 }
 
+data "oci_exec_test" "check-killability" {
+  digest = module.this.image_ref
+  script = "docker kill $(docker run -d --rm $${IMAGE_NAME})"
```
mattmoor, line 48: Are we sure that most containers will start a daemon process properly with the nested command?

mattmoor: Oh, don't mistake my comment for criticism.  I love the idea of stuffing tests here, I am just worried a lot will: print usage and exit.

Maybe a middle ground is to create a common test module for this, and invoke it everywhere we can?

amouat:
Yeah, this sounds good. `docker kill` should work on all images. Point me at a few that fail and I'll take a look and try to make some suggestions.

imjasonh:
> Yeah, this sounds good. `docker kill` should work on all images. Point me at a few that fail and I'll take a look and try to make some suggestions.

[melange](https://github.com/chainguard-images/images/actions/runs/5414071799/jobs/9840635525?pr=996), [apko](https://github.com/chainguard-images/images/actions/runs/5414071799/jobs/9840626571?pr=996) and [ko](https://github.com/chainguard-images/images/actions/runs/5414071799/jobs/9840616272?pr=996) all fail.

amouat:
I guess those are all short-lived containers - I guess this rule only applies to daemon containers :/

I didn't think about that.

amouat:
So with apko etc, the issue is that the default command returns and the container exits before the kill command completes. On my machine it's a race condition - sometimes it works and sometimes it doesn't. We only need a command that runs for a second or so though, so the following should work reliably:

```
docker kill $(docker run -d -v $PWD/examples:/data ghcr.io/wolfi-dev/apko build -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub --arch aarch64 /data/wolfi-base.yaml example out.tar)
```

(Well, with arch set properly).

It's only really important for server applications like redis, nginx etc, but also I suspect some of those might not start as a daemon by default (for example see the node image). In other words, I think this might make a decent "template" test but will need to be modified by a lot of images.

PR #1010, opened 2023-06-29
migrate k8sgpt-operator to TF

imjasonh:


images/k8sgpt/tests/main.tf
```diff
@@ -10,5 +10,5 @@ variable "digest" {
 
 data "oci_exec_test" "runs" {
   digest = var.digest
-  script = "${path.module}/01-runs.sh"
+  script = "docker run --rm $IMAGE_NAME --help"
```
mattmoor, line 13: Hmm, I've been using `$$IMAGE_NAME`, so we may have some cleanup to do

mattmoor (approved):

images/k8sgpt/tests/main.tf
imjasonh: Yep that's right.

PR #1013, opened 2023-06-29
Consolidate a single busybox leg

mattmoor:
We still build it all, this file just controls how `monopod` invokes things and right now this has us invoke the mega module for this target twice instead of once.


joshrwolf (dismissed):

joshrwolf (approved):

mattmoor:
This is failing as HEAD with the same error because of version skew in the Alpine indices, and all this does it make it so we hit it once instead of twice, so bypassing the presubmit check.

PR #1019, opened 2023-06-29
migrate oidc-discovery-provider to spire module

imjasonh:
NB: This actually renames `cgr.dev/chainguard/oidc-discovery-provider` to `cgr.dev/chainguard/spire-oidc-discovery-provider`, which I think is more correct anyway (it's the name of the package, for example).

If this sounds okay, we can see if anybody uses it that needs to be notified of the change.


images/spire/configs/latest.agent.apko.yaml
```diff
@@ -22,4 +22,4 @@ entrypoint:
 annotations:
   "org.opencontainers.image.authors": "Chainguard Team https://www.chainguard.dev/"
   "org.opencontainers.image.url": https://edu.chainguard.dev/chainguard/chainguard-images/reference/spire-agent/
-  "org.opencontainers.image.source": https://github.com/chainguard-images/images/tree/main/images/spire-agent
+  "org.opencontainers.image.source": https://github.com/chainguard-images/images/tree/main/images/spire
```
mattmoor, line 25: JFYI, we've been screwing this up everywhere, and I have a yellow sticky to audit these.

images/spire/main.tf
```diff
@@ -26,7 +26,7 @@ module "latest-dev" {
   for_each = local.components
   source   = "../../tflib/publisher"
 
-  target_repository = var.target_repository
+  target_repository = "${var.target_repository}-${each.key}"
```
mattmoor, line 29: Thanks for fixing this, it was wrong before, but I think it's still wrong for the new component?

images/spire/tests/main.tf
```diff
@@ -46,6 +47,15 @@ resource "helm_release" "spire" {
           tag        = data.oci_string.ref["agent"].pseudo_tag
         }
       }
+      spiffe-oidc-discovery-provider = {
+        enabled = true
+        config  = { acme = { tosAccepted = true } }
+        image = {
+          registry   = data.oci_string.ref["server"].registry
+          repository = data.oci_string.ref["server"].repo
+          tag        = data.oci_string.ref["server"].pseudo_tag
```
mattmoor, line 56: wrong image

mattmoor, line 56:
```suggestion
          registry   = data.oci_string.ref["oidc-discovery-provider"].registry
          repository = data.oci_string.ref["oidc-discovery-provider"].repo
          tag        = data.oci_string.ref["oidc-discovery-provider"].pseudo_tag
```

mattmoor:
This LGTM.  Beware I did push several fixes to your fork, so fetch before you push anything.

We should sort out what sort of traffic this gets and plan a withdraw of the old tags as part of the migration to the new repo.

imjasonh:
Looking through traffic I've only been able to identify our own traffic pushing and pulling the image to test it. Otherwise there are no logged-in users pulling the image, and even anonymous pulls look like our own testing in terms of useragents, timing and volume relative to our pushes.

I believe this change should be safe to make.

images/spire/configs/latest.agent.apko.yaml
imjasonh: Even better, we should be generating these so we can't mess them up.

mattmoor (approved):

PR #1024, opened 2023-06-30
Port external-dns to TF

mattmoor:


imjasonh (dismissed):

mattmoor:
Looks like this just hangs and the `oci_exec_test` is unable to time it out, so it'll be challenging to get any useful output out of it 🤔

mattmoor:
I think @joshrwolf is taking this over

PR #1027, opened 2023-06-30
add docs for building with apko directly

imjasonh:
Also resolve a TODO about documenting k8s testing.


TERRAFORM.md
```diff
@@ -106,4 +106,24 @@ TF_VAR_digest=... terraform apply
...
+The tests use the config file at `$HOME/.kube/config` to connect to the cluster.
+
+### Building with `apko` directly
+
+Images in this repo are designed to be built with Terraform as described above, but they can also be built directly with `apko`:
+
+To build a Wolfi-based image:
+
+```shell
+make cfg=images/static/configs/wolfi.apko.yaml apko-build
```
deitch, line 120: * Do all of the images follow the same pattern? Is it possible to just run `make img=wolfi` or similar?
* We used to use [`generate-matrix` action](https://github.com/chainguard-images/images/blob/main/.github/actions/generate-matrix/action.yml) to generate the list of all images. Does this still work? If not, what is the equivalent for this now?
* Do you have to pick which image type to build? Can it infer it somehow?

imjasonh: Good questions!

> Do all of the images follow the same pattern? Is it possible to just run make img=wolfi or similar?

Mostly, but not reliably enough I think. We can stub out the `images/` prefix and the `.apko.yaml` suffix, but there are lots of differently named YAML files all in the same directory.

> We used to use [generate-matrix action](https://github.com/chainguard-images/images/blob/main/.github/actions/generate-matrix/action.yml) to generate the list of all images. Does this still work? If not, what is the equivalent for this now?

I don't know if we have one now. Any equivalent would have to invoke terraform to get the list of images it built as an output. We can do that though.

> Do you have to pick which image type to build? Can it infer it somehow?

Currently yes. We could built smarts to detect it, but honestly there's only two types, and all but 3 of them are Wolfi-based, so I'd rather not build a ton of smarts just to tell whether it's one of those three.

deitch: > Mostly, but not reliably enough I think. We can stub out the images/ prefix and the .apko.yaml suffix, but there are lots of differently named YAML files all in the same directory.

I think that is why we had that monopod to generate the list. If the images were consistently named, it would be easy enough. IIRC, there were good reasons for it.

> Any equivalent would have to invoke terraform to get the list of images it built as an output. We can do that though.

That might be pretty good. It would replace monopod while it is at it. Anything that lets us invoke some command and get the info we need is good enough.

> We could built smarts to detect it, but honestly there's only two types, and all but 3 of them are Wolfi-based, so I'd rather not build a ton of smarts just to tell whether it's one of those three

I know, and my rational mind agrees, but my irrational instincts worry about anything that has "just these 3 exceptions" growing.

imjasonh: There may be other exceptions in the future, in theory, but it would be very hard to convince me to add a fourth alpine-based image at this point.

mattmoor: Yeah, it would take a pretty significant change for us to consider adding any more non-Chainguard images _**in this repo**_.  I'd be fully supportive of creating them elsewhere with the same toolchain, but here they are tech debt we need to eventually eliminate.

deitch: That does make it somewhat easier. The goals is to be able to build any image using standard CLI tools (i.e. `apko`), so you can do the simple case (not to mention debugging when problems arise). If all images will have identical configs, then it gets much easier.

PR #1030, opened 2023-06-30
Port trust-manager to TF

mattmoor:


images/trust-manager/tests/02-helm-install.sh
```diff
@@ -1,26 +1,7 @@
...
-		exit 1
-	fi
-
-	if [[ "${IMAGE_TAG}" == "" ]]; then
-		echo "Must set IMAGE_TAG environment variable. Exiting."
-		exit 1
-	fi
-}
-
 function ensure_trust_manager(){
```
imjasonh, line 5: We can just do `helm install --wait` and drop the sleep and `kubectl wait`.

Up to you, I can come back and get this later too.

mattmoor: I'll try it now, and if it's not easy we can circle back to it

imjasonh (approved):

PR #1033, opened 2023-06-30
fix calico test, add kube-controllers and calico-cni

imjasonh:
- rename `cgr.dev/chainguard/calico` -> `cgr.dev/chainguard/calicoctl`
- add working test for `calico-node` -- previously the test often didn't test anything because the operator paved over the image update
- add `kube-controllers` and `calico-cni` images, and add to test script


images/calico/configs/latest.node.apko.yaml
```diff
@@ -1,6 +1,7 @@
 contents:
   packages:
     - calico-node
+    - busybox
```
mattmoor, line 4: For posterity, why do we need `busybox`?

images/calico/tests/install.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# monopod:tag:k8s
+
+set -o errexit -o nounset -o errtrace -o pipefail -x
+
+curl -o calico.yaml https://raw.githubusercontent.com/projectcalico/calico/v3.26.1/manifests/calico.yaml
+
+sed -i.bak "s|docker.io/calico/node:.*|${NODE_IMAGE}|g" calico.yaml
+sed -i.bak "s|docker.io/calico/kube-controllers:.*|${KUBE_CONTROLLERS_IMAGE}|g" calico.yaml
+#sed -i.bak "s|docker.io/calico/cni:.*|${CNI_IMAGE}|g" calico.yaml
```
mattmoor, line 11: Do we need this?

imjasonh: Not yet, but that's the next image and I'd like to keep it as a reminder.

images/calico/configs/latest.node.apko.yaml
imjasonh: @joshrwolf knows more, but in short this fails without it. I think the pod does that thing where the binary gets copied out so it needs busybox to do the copying

joshrwolf: the container uses a custom runit script on boot that needs /bin/sh

images/calico/configs/latest.cni.apko.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - username: nonroot
+      uid: 65532
+      gid: 65532
+  run-as: 0
+
+entrypoint:
+  command: /opt/cni/bin/install
+
+environment:
+  PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/cni/bin
```
joshrwolf, line 20: not blocking, just curious for other cni impls we’ll inevitably maintain, was this needed for something else calling the cnis without the full path?

images/calico/tests/install.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# monopod:tag:k8s
```
joshrwolf, line 3:
```suggestion
```

nit

joshrwolf (approved):
just a few nits, otherwise lgtm for this iteration with out current test harness

images/calico/configs/latest.cni.apko.yaml
imjasonh: This wasn't needed to make the tests pass, but this is consistent with the upstream image is all.

PR #1035, opened 2023-06-30
Drop non-terraform support, including k8s testing

mattmoor:


imjasonh (approved):

mattmoor:
Ok, so there are 4 failing legs here (and the roundup), triaging:
 - `busybox` and `git` are toast until Alpine serves consistent indices.
 - `hugo` is broken because we inadvertently dropped the `dev` packages from the Wolfi `dev` image (Fixed here: https://github.com/chainguard-images/images/pull/1071)
 - `buck2` appears broken at HEAD by some starlark thing, started a slack thread

I'm inclined to bypass presubmits and merge since none of these issues are related to this change.

imjasonh (approved):

PR #1036, opened 2023-06-30
Remove unused scripts

mattmoor:


imjasonh (approved):

mattmoor:
I'm not sure why this ran all the tests, but I think these are all problems at HEAD since none are in this directory, and several of them are "usual suspects".

PR #1051, opened 2023-07-05
support -var=archs to specify only certain archs to build for

imjasonh:


main.tf
```diff
@@ -23,11 +23,16 @@ variable "extra_packages" {
   default = []
 }
 
+variable "archs" {
+  type    = list(string)
+  default = []
```
joshrwolf, line 28: how about just setting this? then we don't need the ternary below

```suggestion
  default = ["x86_64", "aarch64"]
```

imjasonh: Good call, I just did `archs` and `alpine_archs` and gave them defaults.

TERRAFORM.md
```diff
@@ -84,6 +84,17 @@ TF_VAR_target_repository=... terraform apply \
...
 ```
 
+### Building for certain architectures
+
+By default all Wolfi-based images are built for `x86_64` (`amd64`) and `aarch64` (`arm64`),
+and Alpine-based images are built for all architectures supported by Alpine.
+
+During testing it can be useful to only build for certain architectures, such as the host architecture.
+
+To achieve this, pass `-var='archs=["x86_64"]'` or `-var='archs=["aarch64"]'` to `terraform apply`.
```
jdolitsky, line 94: can we also use `TF_VAR_archs='["x86_64"]'` ?

jdolitsky (approved):

PR #1058, opened 2023-07-05
document building from local packages

imjasonh:
cc @strongjz


rawlingsj:
I tried the local packages docs here but ran into an issue. It may well be my setup but the error was

```
│ Error: Invalid index
│
│   on tflib/version-tags/main.tf line 13, in output "tag_list":
│   13:   ][0][0]
│     ├────────────────
│     │ var.config.contents.packages is list of string with 24 elements
│     │ var.package is "kube-fluentd-operator"
│
│ The given key does not identify an element in this collection value: the collection has no elements.
╵
╷
│ Error: Invalid index
│
│   on .terraform/modules/kube-fluentd-operator.latest.this/main.tf line 75, in resource "cosign_attest" "slsa-provenance":
│   75:         for k in data.apko_config.this.config.contents.packages : split("=", k)[0] => split("=", k)[1]
│
│ The given key does not identify an element in this collection value: the given index is greater than or equal to the length of the collection.
```

the main.tf for my `kube-fluentd-operator`  module has this 

```
module "version-tags" {
  source  = "../../tflib/version-tags"
  package = "kube-fluentd-operator"
  config  = module.latest.config
}
```

imjasonh:
> I tried the local packages docs here but ran into an issue. It may well be my setup but the error was

This error indicates that one of the packages in image had an inconsistent version between the two architectures. This is definitely not clear from the error message, so we should both a.) improve the error message, and b.) document it here in TERRAFORM.md.

But at least this isn't an indication that the build instructions in general aren't right, just that you hit a weird temporary case that we're working to fix.

Thanks for testing it out!

rawlingsj:
Ah that explains it - thanks.  Yeah it started working as soon as I specified the arch, I'm guessing because when I copy my local `./packages` folder it only contains arm apks and not x86.

One bit of feedback - I found that when I build a local apk and copy the `./packages` folder to the root of my images fork to test... terraform doesn't recognise that there's a new apk and uses the previously built image.  For a while I thought I was testing my latest fixes but an old image was being deployed via the test.

My workaround is to add an env var to my apko config and bump the value each time I copy new packages.  I'm sure there's a better terraform approach to do this.

rawlingsj (approved):

PR #1060, opened 2023-07-05
fix up and enable k8s CSI tests

imjasonh:
- remove version tests
- remove dev tests
- streamline deploy tests and rename to `deploy.sh`

In the future we might want to consolidate some of these into a module together, but for now at least running the tests is nice.


images/kubernetes-csi-external-provisioner/tests/deploy.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# monopod:tag:k8s
```
jonjohnsonjr, line 3: This still a thing?

imjasonh: It is not!

jonjohnsonjr (approved):

PR #1067, opened 2023-07-06
WIP: real GCP test for secrets-store-csi-driver-provider-gcp

imjasonh:
This creates a GKE cluster (non-autopilot*), creates a GCP secret with secret data, and installs CSI secrets-store provider and (currently) the _upstream_ provider image, runs a secret that logs the secret data, and checks that the logs contain that secret data.

The pod currently doesn't start due to

```
  Warning  FailedMount  31s (x9 over 12m)   kubelet            MountVolume.SetUp failed for volume "secret" : rpc error: code = Unknown desc = failed to mount secrets store objects for pod default/test-pod-driving-sunbeam, err: rpc error: code = PermissionDenied desc = unable to obtain auth for mount: unable to obtain workload identity auth: unable to fetch identitybindingtoken: could not get idbindtoken token, status: 404
```

Once this works though, we should be able to patch in our own GCP provider image (and k8s CSI image if we want) to get pre-release validation of the image in a real cluster before we tag it.

Open questions include:
- do we want to run this (and how/when) when building the image -- should it be on a different cadence, with real GCP creds to create the cluster?

\* autopilot doesn't let us install CSI drivers, GCP secret access is managed by GCP in autopilot-land.


imjasonh:
This is not currently wired into the root main.tf, but it can be run manually on-demand. I'd like to merge this to save progress, while we figure out how we want to wire this up as a real test.

imjasonh:
> Can we at least at a README for now that includes two things:

Done!

joshrwolf (approved):
lgtm! only comment is we hoist the infra building stuff (`gke`) into a module we can use elsewhere, but that can happen in a follow up

PR #1070, opened 2023-07-06
build calico/pod2daemon-flexvol image

imjasonh:


images/calico/configs/latest.pod2daemon.apko.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+accounts:
+  groups:
+    - groupname: nonroot
+      gid: 65532
+  users:
+    - username: nonroot
+      uid: 65532
+      gid: 65532
+  run-as: 0
```
mattmoor, line 14: intentional?

imjasonh: Yeah, I think this is necessary. The upstream image runs as root and the entrypoint copies its binary out into another location so it can be used by another pod.

PR #1079, opened 2023-07-07
remove unnecessary test-foo-dev modules

imjasonh:
We keep some dev tests that test meaningful things about the dev variant, otherwise testing the dev image doesn't really provide much value, and makes tests slower and flakier.


images/argocd/main.tf
```diff
@@ -44,8 +44,6 @@ module "version-tags" {
 }
 
 module "test-latest" {
-  for_each = local.components
```
mattmoor, line 47: ouch, I bet that speeds things up!

mattmoor (approved):

PR #1081, opened 2023-07-07
Mount the host's cache directory into the container's reproducibility check

mattmoor:
This should improve performance, but it's really a reliability change because we periodically see network flakes downloading APKs, and given the nature of this test we shouldn't ever need to download anything if we have a hot cache (which we should!).


imjasonh (approved):

mattmoor:
The one failing leg here is failing at HEAD and Jason has a fix here, so bypassing presubmits to land this: https://github.com/chainguard-images/images/pull/1082

Issue #941, opened 2023-06-26
policy-check-image outdated tagging logic


joshrwolf:
the `policy-check-image` action (run postsubmit on the `release` workflow) is expecting an outdated tagging methodology.

with the addition of "grouped" images (#897), `policies/main.tf` is currently wired up to expect the tags to be:

```
cgr.dev/chainguard/certmanager:latest.webhook
```

when it should actually be:

```
cgr.dev/chainguard/certmanager-webhook:latest
```

for now, we'll disable the `policy-check-action` legs in the `release` workflow, but will revisit when the toil of the terraform migration has subsided.

PR #1089, opened 2023-07-10
remove policy-check-image

imjasonh:
Fixes https://github.com/chainguard-images/images/issues/941

We're going to move this into TF as a post-build-pre-tag check, so we can ensure this for all variants. If we want to check this in the local workflow it means we'll have to sign in the local workflow.


mattmoor:
I don't think we want to close the issue yet.

mattmoor (approved):

PR #1113, opened 2023-07-13
Calico cni tests

joshrwolf:
I cannot stress enough how much I dislike the k8s related terraform provider landscape, but this seems to work 🤷


images/calico/kind-tests/main.tf
```diff
@@ -0,0 +1,99 @@
...
+  kind_config {
+    kind        = "Cluster"
+    api_version = "kind.x-k8s.io/v1alpha4"
+    networking {
+      disable_default_cni = true
+    }
+  }
+}
+
+data "oci_exec_test" "smoke" {
```
imjasonh, line 95: Let's make a `var` for this image, with the default being `:latest`, and pass it through an `oci_ref` to resolve it to digest before passing here.

Then we can have a simple repro of `:latest` with the opportunity to override it with whatever other image we might be testing.

imjasonh (commented):
This broadly looks good. I'd love to have a README to help walk me through trying it out myself.

Thanks for adding this!

images/calico/kind-tests/main.tf
joshrwolf: great idea! pushed with latest changes that actually use our images 😅

joshrwolf:
this will fail until https://github.com/wolfi-dev/os/pull/3631 is built

imjasonh (approved):

PR #1116, opened 2023-07-14
Fixing php-fpm image and test

erikaheidi:
This PR fixes the php-fpm image entrypoint - _now **for real**_ - and adds a more robust test to check that the service is really up and running.

### DIFF
--- a/images/php/tests/main.tf
+++ b/images/php/tests/main.tf
@@ -32,5 +32,5 @@ data "oci_exec_test" "composer" {
 data "oci_exec_test" "fpm" {
   count  = var.check-fpm ? 1 : 0
   digest = var.digest
-  script = "docker run --rm $IMAGE_NAME -v"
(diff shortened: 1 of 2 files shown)


rawlingsj (approved):

images/php/tests/main.tf
```diff
@@ -32,5 +32,5 @@ data "oci_exec_test" "composer" {
 data "oci_exec_test" "fpm" {
   count  = var.check-fpm ? 1 : 0
   digest = var.digest
-  script = "docker run --rm $IMAGE_NAME -v"
+  script = "FPMRUN=$( timeout --preserve-status -s TERM 4s docker run --rm $IMAGE_NAME 2>&1 ) && echo $FPMRUN | grep -q 'ready to handle connections'"
```
rawlingsj, line 35: just a nit - it might be easier to maintain if this was moved into a bash script, as then we could run it locally without applying the terraform.  Not blocking the merge, just a thought for the future.

rawlingsj: going to resolve the comment as it is non blocking

PR #1126, opened 2023-07-17
Update ebs csi driver with mount package

strongjz:
Add mount package to image

### DIFF
--- a/images/aws-ebs-csi-driver/configs/latest.apko.yaml
+++ b/images/aws-ebs-csi-driver/configs/latest.apko.yaml
@@ -1,7 +1,7 @@
 contents:
   packages:
     - aws-ebs-csi-driver
-    - mount
+    - umount
 
 accounts:
   groups:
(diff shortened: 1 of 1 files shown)


images/aws-ebs-csi-driver/configs/latest.apko.yaml
```diff
@@ -1,6 +1,7 @@
 contents:
   packages:
     - aws-ebs-csi-driver
+    - mount
```
imjasonh, line 4: I think we need `umount`, the error we saw was

```
unmount failed: exec: "umount": executable file not found in $PATH
```

imjasonh, line 4:
```suggestion
    - umount
```

imjasonh (approved):
This is definitely telling us we need better test coverage for this image. Which #1090 should give us!

PR #1128, opened 2023-07-18
add tekton pipeline+chains images

imjasonh:
cc @wlynch @strongjz


images/tekton/tests/test.sh
```diff
@@ -0,0 +1,58 @@
...
+CONTROLLER_IMAGE=${CONTROLLER_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/controller:v0.49.0}
+ENTRYPOINT_IMAGE=${ENTRYPOINT_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/entrypoint:v0.49.0}
+EVENTS_IMAGE=${EVENTS_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/events:v0.49.0}
+NOP_IMAGE=${NOP_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/nop:v0.49.0}
+RESOLVERS_IMAGE=${RESOLVERS_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/resolvers:v0.49.0}
+SIDECARLOGRESULTS_IMAGE=${SIDECARLOGRESULTS_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/sidecarlogresults:v0.49.0}
+WEBHOOK_IMAGE=${WEBHOOK_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/webhook:v0.49.0}
+WORKINGDIRINIT_IMAGE=${WORKINGDIRINIT_IMAGE:-gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/workingdirinit:v0.49.0}
+
+curl -sL https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml | \
```
mattmoor, line 14: Should "latest" here be `v0.49.0`?

images/tekton/tests/test.sh
```diff
@@ -0,0 +1,58 @@
...
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: test-taskrun
+spec:
+  taskSpec:
+    steps:
+      - image: cgr.dev/chainguard/busybox
+        script: echo "hello world"
+EOF
```
mattmoor, line 49: This feels like a good opportunity to test a `tkn` image :)

images/tekton/tests/test.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Wait for up to 5 minutes to see the TaskRun complete successfully.
```
mattmoor, line 51: this is actually longer than the default `oci_exec_test` timeout, so maybe set that so it's clearer?

images/tekton/tests/test.sh
```diff
@@ -0,0 +1,58 @@
...
+
+curl -sL https://storage.googleapis.com/tekton-releases/chains/latest/release.yaml | \
+    sed "s|gcr.io/tekton-releases/github.com/tektoncd/chains/cmd/controller[a-z0-9:@.]\{1,\}|${CHAINS_IMAGE}|g" | \
+    kubectl apply -f -
+
+sleep 5
+
+kubectl wait --for=condition=Ready pods --all -n tekton-pipelines --timeout=300s
+kubectl wait --for=condition=Ready pods --all -n tekton-pipelines-resolvers --timeout=300s
+kubectl wait --for=condition=Ready pods --all -n tekton-chains --timeout=300s
```
mattmoor, line 35: We aren't actually testing the chains functionality.

images/tekton/tests/main.tf
```diff
@@ -0,0 +1,68 @@
...
+  }
+  env {
+    name  = "WEBHOOK_IMAGE"
+    value = var.digests["webhook"]
+  }
+  env {
+    name  = "WORKINGDIRINIT_IMAGE"
+    value = var.digests["workingdirinit"]
+  }
+
```
mattmoor, line 67:
```suggestion
```

images/tekton/tests/test.sh
imjasonh: Shortened the timeout anyway.

images/tekton/tests/test.sh
imjasonh: Getting chains set up and configured and testing it seemed like it would be a diminishing return. I'm fairly confident that if we're packaging the same upstream code into an image and it comes up in the cluster, there's not really much else we can get wrong about it that a bigger test would catch.

The chains controller is based on our nonroot static image already, for instance.

PR #1132, opened 2023-07-19
define standard annotations in TF

imjasonh:
This starts to let TF module invocations fill in all annotations, which we can then remove from the YAML config.

As an example, `crane`'s YAML config no longer specifies annotations, but its invocation of the tflib/publisher module includes a new `name` field based on `basename(path.module)` (which is `crane`).

This doesn't remove copypasta, but it removes the need to copypasta-and-remember-to-update-a-string in each YAML config.

```
$ terraform plan -target=module.crane -var=target_repository=ttl.sh/jason
+ annotations = {
    + "org.opencontainers.image.authors" = "Chainguard Team https://www.chainguard.dev/"
    + "org.opencontainers.image.source"  = "https://github.com/chainguard-images/images/images/crane"
    + "org.opencontainers.image.url"     = "https://edu.chainguard.dev/chainguard/chainguard-images/reference/crane/"
}
```

Modules that don't specify tflib/publisher's `name` still get the value from the YAML:

```
$ terraform plan -target=module.zookeeper -var=target_repository=ttl.sh/jason
+ annotations = {
    + "org.opencontainers.image.authors" = "Chainguard Team https://www.chainguard.dev/"
    + "org.opencontainers.image.source"  = "https://github.com/chainguard-images/images/tree/main/images/zookeeper"
    + "org.opencontainers.image.url"     = "https://edu.chainguard.dev/chainguard/chainguard-images/reference/zookeeper/"
}
```

(This makes it easier to roll out progressively, and doesn't punish in-flight PRs doing it the old way)

This change also adds validation before running the structure test that these annotations are set, so we don't forget to set `name`.

If this looks good we can drop `default_annotations` from the publisher module (https://github.com/chainguard-dev/terraform-publisher-apko/pull/16).


tflib/publisher/main.tf
```diff
@@ -24,12 +24,33 @@ variable "extra_packages" {
...
+  default = "TODO" // TODO remove this
+}
+
+output "path" {
+  value = basename(path.cwd)
+}
+
+locals {
+  updated_config = merge(yamldecode(var.config),
+    var.name == "TODO" ? {} :
```
mattmoor, line 38: what isn't migrated, or is this for merge conflicts?

tflib/publisher/main.tf
```diff
@@ -24,12 +24,33 @@ variable "extra_packages" {
...
+  )
+}
+
 module "this" {
   source  = "chainguard-dev/apko/publisher"
   version = "0.0.6"
 
   target_repository = var.target_repository
-  config            = var.config
+  config            = yamlencode(local.updated_config)
```
mattmoor, line 53: this surgery is gross, can we cut a release of the module with your change, bump the version here and avoid the roundtrip?

tflib/publisher/main.tf
imjasonh: This is mainly here for in-flight PRs that do it the old way, so we can clean those up in a post-merge pass.

tflib/publisher/main.tf
imjasonh: We can, sure. I kinda thought it was nice not to do this surgery in the provider, where we need releases to change the logic, but I get that moving more stuff into HCL is not necessarily a win.

mattmoor: I think for annotations the provider already does this surgery, right?

mattmoor: Given the size of the change, I'd also be ok with checkpointing this and cleaning it up with the release and dropping the TODOs in follow-up.

mattmoor (approved):

PR #1144, opened 2023-07-26
Group minio and minio-client into same directory

tcnghia:
This follows the convention for Kyverno and Tekton, which allows us to test both of these images together.


images/minio/main.tf
```diff
@@ -8,47 +8,59 @@ variable "target_repository" {
...
 module "latest" {
-  source = "../../tflib/publisher"
+  for_each = local.packages
+  source   = "../../tflib/publisher"
 
   name = basename(path.module)
 
-  target_repository = var.target_repository
-  config            = file("${path.module}/configs/latest.apko.yaml")
+  target_repository = "${var.target_repository}${trimprefix(each.key, "minio")}"
```
mattmoor, line 24: look at what kyverno does here.

images/minio/main.tf
```diff
@@ -8,47 +8,59 @@ variable "target_repository" {
...
+  package  = each.value
+  config   = module.latest[each.key].config
 }
 
 module "test-latest" {
-  source = "./tests"
-  digest = module.latest.image_ref
+  for_each = local.packages
+  source   = "./tests"
+  digest   = module.latest[each.key].image_ref
```
mattmoor, line 53: We should make this pass all of the constituent images as `digests = { ... }` and then we can run the current `--version` test on each, which will make the follow up to test them against each other much more contained.

tcnghia: Done

images/minio/main.tf
tcnghia: Done

images/minio/tests/main.tf
```diff
@@ -4,11 +4,13 @@ terraform {
   }
 }
 
-variable "digest" {
-  description = "The image digest to run tests over."
+variable "digests" {
+  description = "The image digests to run tests over."
```
mattmoor, line 8: generally this enumerates the keys in a type, check one of the example images.

tcnghia: Done.

mattmoor (approved):

PR #1150, opened 2023-07-27
Consolidate the JRE testing

mattmoor:
This removes the duplicate hello-world test and has the JDK run all of the JRE testing, but using itself as the `sdk-image`


images/jdk/tests/main.tf
```diff
@@ -13,7 +13,8 @@ data "oci_exec_test" "version" {
   script = "docker run --rm $IMAGE_NAME javac -version"
 }
 
-data "oci_exec_test" "hello-world" {
-  digest = var.digest
-  script = "${path.module}/02-hello-world.sh"
+module "jre-test" {
+  source    = "../../jre/tests"
+  digest    = var.digest
+  sdk-image = var.digest
```
jonjohnsonjr, line 19:
```suggestion
  sdk-image = var.sdk-image
```

jonjohnsonjr: Should the tests have passed like this?

mattmoor: Yes because you are incorrect.  The JDK is both the runtime and the SDK here

jonjohnsonjr (approved):

PR #1152, opened 2023-07-28
Add a hello-world ruby test

tcnghia:


images/ruby/tests/main.tf
```diff
@@ -17,3 +17,8 @@ data "oci_exec_test" "irb-version" {
   digest = var.digest
   script = "docker run --rm --entrypoint irb $IMAGE_NAME -v"
 }
+
+data "oci_exec_test" "hello-world" {
+  digest = var.digest
+  script = "${path.module}/02-hello-world.sh"
+}
```
mattmoor, line 24: This is good, but we should also add testing that exercises the `gem` path.

tcnghia: Done

ajayk (approved):

PR #1155, opened 2023-07-28
Add support for earlier JREs

mattmoor:


ajayk (commented):
do we need the jdk8.tags file

ajayk (approved):

PR #1156, opened 2023-07-28
add 'go test ./...' able e2e tests for eks components

joshrwolf:
This plays with a new framework for e2e tests that require significant dependent infrastructure to sufficiently test.

This demonstrates e2e tests for EKS components (`aws-load-balancer-controller` and `aws-ebs-csi-driver`). 

It uses `go` anad `terratest` to do the light orchestration of terraform apply's to workaround the well known limitations of mixing infra + k8s into the same tf apply, while feeling pretty ergonomic on the go side of the house.

The tests themselves can be kicked off with:

```
cd images/tests/eks

# go test defaults to 10m
go test -timeout 20m -v -tags=e2e ./...
```

This PR sets up a go.mod root at `images/go.{mod,sum}`, so it stands to reason in the future we could just run all the e2e tests with the appropriate `go test ... ./...`.

I'm not attached to the folder structure, but I did end up liking how `e2e_test.go` turned out, and the testing iteration felt very nice.

CI is all sorts of busted, but I'll address those once the more interesting details are hashed out.


images/tests/eks/e2e_test.go
```diff
@@ -0,0 +1,92 @@
...
+				},
+			}
+
+			down, err := tUp(t, opts)
+			if err != nil {
+				t.Fatalf("Failed to create ALB: %s", err)
+			}
+			defer down()
+
+			t.Run("should have a valid ALB", func(t *testing.T) {
```
joshrwolf, line 52: I like that we can get as complicated as we want here, either with go, or with more terraform orchestrated through terratest

images/tests/eks/e2e_test.go
```diff
@@ -0,0 +1,94 @@
...
+		opts := terraform.WithDefaultRetryableErrors(t, &terraform.Options{
+			TerraformDir: setupPath,
+		})
+		ts.SaveTerraformOptions(t, setupPath, opts)
+		terraform.InitAndApply(t, opts)
+	})
+	setupOpts := ts.LoadTerraformOptions(t, setupPath)
+
+	// Group tests that require an EKS cluster to exists
+	t.Run("eks", func(t *testing.T) {
```
joshrwolf, line 31: seems like a nice grouping of all the `eks` components we want to test that can share the same setup.

this can be adopted in a similar way for things like common `gcp` or `az` setup

PR #1159, opened 2023-07-29
Expand the list of images tested with prometheus helm chart

mattmoor:
This deploys every image I could find referenced from the `kube-prometheus-stack` `values.yaml`, and I confirmed locally that things like `alertmanager` and `prometheus` spin up with this configuration, and that it is using our images.

I also made it use our public KSM image (and we should consider doing the same in the KSM test that spins up this same Helm chart).

However, despite standing up more stuff, this doesn't actually functionally test them still, but it should enable us to follow up and test things.  The one thing that IS tested is the `node-exporter` where I changed the test to hit the DaemonSet Service created by the Helm chart instead of spinning up its own pod.

### DIFF
--- a/images/prometheus/tests/main.tf
+++ b/images/prometheus/tests/main.tf
@@ -45,12 +45,6 @@ data "oci_exec_test" "cloudwatch-runs" {
   working_dir = path.module
 }
 
-data "oci_exec_test" "node-runs" {
-  digest      = var.digests["node-exporter"]
-  script      = "./node-runs.sh"
-  working_dir = path.module
-}
-
 data "oci_exec_test" "redis-runs" {
   digest      = var.digests["redis-exporter"]
   script      = "./redis-installs.sh"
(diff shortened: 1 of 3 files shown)


images/prometheus/tests/main.tf
```diff
@@ -78,7 +78,48 @@ resource "helm_release" "kube-prometheus-stack" {
...
+      alertmanager = {
+        enabled = true
+        alertmanagerSpec = {
+          image = {
+            registry   = data.oci_string.ref["alertmanager"].registry,
+            repository = data.oci_string.ref["alertmanager"].repo,
+            tag        = data.oci_string.ref["alertmanager"].pseudo_tag,
+          }
+        }
+      }
```
mattmoor, line 100: I'm not seeing these actually deploy.

I think we may want to deviate from our "generate values.yaml" approach for this one because the upstream version is 4000 lines long.

jdolitsky (approved):

PR #1162, opened 2023-07-31
Add an assertion to catch k3s tagging woes

mattmoor:
Drop `APKO_IMAGE` echo to avoid local dev annoyance.  This was for bring-up to make sure it was coming through from the runner.

### DIFF
--- a/images/k3s/main.tf
+++ b/images/k3s/main.tf
@@ -75,8 +75,32 @@ module "tagger" {
   tags = merge(
     { for t in toset(concat(["latest"], module.version-tags.tag_list)) : t => module.latest.image_ref },
     { for t in toset(concat(["latest"], module.version-tags.tag_list)) : "${t}-dev" => module.latest-dev.image_ref },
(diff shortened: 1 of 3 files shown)


images/k3s/main.tf
```diff
@@ -75,8 +75,32 @@ module "tagger" {
...
+  source = "../../tflib/tagger"
+
+  depends_on = [
+    module.test-latest,
+    module.test-embedded,
+  ]
+
+  tags = merge(
+    { for t in toset(concat(["latest"], module.version-tags.tag_list)) : "${t}" => module.latest-images.image_ref },
+    { for t in toset(concat(["latest"], module.version-tags.tag_list)) : "${t}-dev" => module.latest-images-dev.image_ref },
```
joshrwolf, line 91:
```suggestion
```

just a nit

joshrwolf (approved):

images/k3s/main.tf
mattmoor: discussed offline, this is consistent, but worth cleaning up more completely in follow-up.

PR #1171, opened 2023-08-02
K3s aio

joshrwolf:
follow up to, and depends on: https://github.com/wolfi-dev/os/pull/4106

this removes the confusing `k3s-embedded` image (details in ^), and removes the poorly named `k3s-images` in favor of `k3s-aio`.

inlines the description for what `k3s-aio` is in the image's config

re: `aio` naming. I'm open to suggestions if this doesn't seem descriptive enough! there's no "upstream" verbiage to go off of here so I made something up.


joshrwolf:
> My only naming suggestion is "allinone" is more descriptive than "aio" (if that is indeed what it stands for)

done! ptal

imjasonh (approved):

PR #1192, opened 2023-08-09
withdraw removed k3s images

joshrwolf:
withdraws the image variants removed as part of #1171


joshrwolf:
obsolete with the great withdraw of 2023

PR #1193, opened 2023-08-09
WIP: split alpine-based image builds

imjasonh:
We have a handful of Alpine-based images for legacy reasons:

- `cgr.dev/chainguard/static:latest`
- `cgr.dev/chainguard/busybox:latest` and `:<version>`, with the `-dev` variant
- `cgr.dev/chainguard/git:latest` and `:<version>`, with the `-root` and `-dev` variants

The Wolfi-based alternatives `:latest-glibc` or `:glibc-<version>(-root)(-dev)`.

Previously we built the Alpine-based images during the same release process as our Wolfi-based ones, but since the Alpine-based images build for many more architectures, and since those builds for those architectures finish at different times, we've seen a lot of errors around inconsistent package versions, most lately for `git`:

```
    "summary": "Unable to lock package \"git-lfs\" to a consistent version",
    "detail": "3.3.0-r3 (riscv64), 3.3.0-r4 (386, amd64, arm/v6, arm/v7, arm64, ppc64le, s390x)",
```

This PR starts to move us toward having a separate release workflow for Alpine-based images, so that we can run it less often, and handle failures more gracefully.

WIP:
- this doesn't change anything about our GH release workflow yet, so as-is this will just stop rebuilding Alpine-based images

Considerations:
- the signer identity still needs to be the same release.yaml workflow, or else verifiers may fail their policies.


.github/workflows/presubmit-build.yaml
```diff
@@ -27,6 +27,20 @@ jobs:
           sharding-factor: 1
           modified-files: ${{ steps.files.outputs.all_changed_files }}
 
+  presubmit-alpine:
```
joshrwolf, line 30: would it be useful to add some `change` checks on these alpine legs on `./alpine`? we could benefit a lot from less potential flakes

joshrwolf:
just some naming nits not worth mentioning, but otherwise I'm all for this 🎉

PR #1195, opened 2023-08-09
add hcl generator initially scoped for output generation

joshrwolf:
creates a **super simple** cli that parses `./main.tf` for module invocations and creates an appropriate `./outputs.tf`

the workflow for a new image (`foo`) becomes:

1. create `images/foo/main.tf`
2. ensure `output.images` exists (similar requirement as `tagger`)
3. run the generator (`./generator`)
4. ???
5. profit

This intentionally leaves the image modules ouptut a manual process. Since it depending on each module, the produced images can vary wildly. The 3 common approaches are outlined in the PR.


images/flux/main.tf
```diff
@@ -90,3 +90,15 @@ module "tagger" {
...
+# for_each to produce a map of each images required attributes.
+output "images" {
+  value = {
+    for k, v in module.latest : k => {
+      oci_ref     = v.image_ref
+      oci_tags    = module.tagger[k].oci_tags
+      apko_config = v.config
+    }
+  }
+}
```
joshrwolf, line 104: This is where things get hairy and why these aren't autogenerated. This is a dynamic traversal that rolls all the way up through builtin terraform commands (`merge`), that require the terraform runtime, rather than the simple `hcl` parser functions we leverage.

Examples like these are ultimately why we just let the user create this output format, similar to how we already make the user create the tags.

PR #1202, opened 2023-08-11
update otelc test cases

joshrwolf:
Expanded the test sweeps we run for `otelc` into 2 groups (deployment and daemonset). Also turn on more receivers/exporters to allow for testing more things.

While many of these settings don't trigger an unhealthy startup, some spew actionable errors that we should be able to catch now with the future e2e tool.

Also adds `busybox` to the image to allow for configurations using certain receivers (like `eks`) to scrape host data, preventing errors like this:

```
2023-08-11T01:26:47.344Z    info    gohai/gohai.go:54    Failed to retrieve platform metadata    {"kind": "exporter", "data_type": "metrics", "name": "datadog", "error": " exec: \"uname\": executable file not found in $PATH"}                                                                                                                       
```

### DIFF
--- a/images/opentelemetry-collector-contrib/configs/latest.apko.yaml
+++ b/images/opentelemetry-collector-contrib/configs/latest.apko.yaml
@@ -2,6 +2,8 @@ contents:
   packages:
     - opentelemetry-collector-contrib
     - opentelemetry-collector-contrib-compat
+    # Required for hostmetrics receivers and certain resourcedetection processors (like eks)
+    - busybox
 
 accounts:
   groups:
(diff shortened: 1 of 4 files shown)


images/opentelemetry-collector-contrib/configs/latest.apko.yaml
```diff
@@ -2,6 +2,8 @@ contents:
   packages:
     - opentelemetry-collector-contrib
     - opentelemetry-collector-contrib-compat
+    # Required for hostmetrics receivers and certain resourcedetection processors (like eks)
+    - busybox
```
joshrwolf, line 6: this adds a shell which we don't really need. we really just need a few utils here (I think just `uname`)? but we don't individually package `uname` so er on the side of caution for now.

rawlingsj (approved):

PR #1272, opened 2023-08-29
fix: add redis-cli as it is required by the tests

ajayk:
- redis moved to sub-packages ,  `redis-cli` is not included in the main package any more

### DIFF
--- a/images/redis/configs/latest.apko.yaml
+++ b/images/redis/configs/latest.apko.yaml
@@ -1,6 +1,7 @@
 contents:
   packages:
     - redis
+    - redis-cli
     - busybox
 
 accounts:


images/redis/configs/latest.apko.yaml
```diff
@@ -1,6 +1,7 @@
 contents:
   packages:
     - redis
+    - redis-cli
```
mattmoor, line 4: I think we want your fix from downstream, I'll stage a PR that does that.

mattmoor: https://github.com/chainguard-images/images/pull/1287

PR #1279, opened 2023-08-29
Add test that tigera-operator can use our calico images

imjasonh:
- add a test that Tigera can be configured to run our Calico images
- add docs to README to describe how

Depends on #1290


imjasonh:
Superceded by #1292

PR #1285, opened 2023-08-30
upgrade kind k8s version to 1.25 to fix Tekton tests

imjasonh:
Tests fail with:

```
"Version check failed","commit":"2299b15","error":"kubernetes version \"1.24.15\" is not compatible, need at least \"1.25.0-0\" (this can be overridden with the env var \"KUBERNETES_MIN_VERSION\")"
...
```


mattmoor (approved):
deja vu

PR #1296, opened 2023-08-31
Migrate our JDK images to follow #1172

mattmoor:
This starts to reduce our reliance on version tags and eliminate boilerplate.


images/jdk/config/main.tf
```diff
@@ -0,0 +1,32 @@
...
+
+data "apko_config" "this" {
+  config_contents = file("${path.module}/template.apko.yaml")
+  extra_packages  = var.extra_packages
+}
+
+output "config" {
+  value = jsonencode(merge(data.apko_config.this.config, {
+    "environment" = {
+      # TODO: Check that this preserves LANG from the template or move that here.
```
mattmoor, line 28: I still need to confirm this and remove the comment or fix.  DO NOT SUBMIT

imjasonh (approved):

images/jdk/config/main.tf
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+
+# TODO: We should have virtual packages that we use here
+# that the appropriate versions "provide".
+
+variable "extra_packages" {
+  description = "The additional packages to install (e.g. openjdk-17)."
+  default     = ["openjdk-17", "openjdk-17-default-jvm"]
+}
```
mattmoor, line 13: @kaniini this is where I'd love to have something unversioned for JDK and have LTS variants.

The "right" default is not actually `default-jvm` because that gets us `openjdk-20-default-jvm`.

I think ideally we would have a way to say: `default-{jdk,jvm}[-lts]`

Then I think our goal state should be:
 - `cgr.dev/chainguard/jre` -> `default-jvm`
 - `cgr.dev/chainguard/jre-lts` -> `default-jvm-lts`
 - `cgr.dev/chainguard/jdk` -> `default-jdk`
 - `cgr.dev/chainguard/jdk-lts` -> `default-jdk-lts`

> Aside: I'm aware that this would be a breaking change for `jre` and `jdk`, but we could introduce `{jre,jdk}-lts`, send notifications, and then shift `{jre,jdk}` to track latest (vs. LTS)

PR #1304, opened 2023-09-01
Migrate our Python images to follow #1172

mattmoor:


images/python/main.tf
```diff
@@ -1,29 +1,73 @@
...
+  name = basename(path.module)
+
+  target_repository = var.target_repository
+  # Make the dev variant an explicit extension of the
+  # locked original.
+  config = jsonencode(module.latest.config)
+  extra_packages = concat(module.dev.extra_packages, [
+    "build-base",
+    "py${local.version}-pip",
+    "python-${local.version}-dev"
```
mattmoor, line 47: We [have](https://github.com/wolfi-dev/os/blob/617c5cb8acc5de499d9178f12c88bd117043c3db/python-3.11.yaml#L10-L11) a `python-3` virtual, and we [have](https://github.com/wolfi-dev/os/blob/617c5cb8acc5de499d9178f12c88bd117043c3db/python-3.11.yaml#L102) a `python3-dev` virtual.

However, we do not have one for `py3-pip` which we would need to make our public images strictly track virtuals, so we can get rid of the explicit `3.11` we have checked in here and rely exclusively on the APK version resolution logic.

Is this as simple as creating a `py3-pip` virtual that these pip packages provide?

mattmoor: Bah, we do have `py3-pip`!

mattmoor: https://github.com/wolfi-dev/os/blob/617c5cb8acc5de499d9178f12c88bd117043c3db/py3.11-pip.yaml#L11C9-L11C16

mattmoor: I think I missed this because we are pretty inconsistent about applying `provides` to older versions of things, e.g. `py3.10-pip` doesn't have a `provides` for `py3-pip`, which may have been what I was looking at.

mattmoor: cool that seems to pass.

dlorenc (approved):

PR #1310, opened 2023-09-02
Migrate our Consul image to follow #1172

mattmoor:


dlorenc (approved):

mattmoor:
I suspect that I may need to modify the helm test, so that we can support testing multiple versions downstream, but I'm waiting to see if is fails first 🤞

mattmoor:
Yeah, I see this in my downstream tests with two versions:
```
Warning: Helm release "consul" was created but has a failed status. Use the `helm` command to investigate the error, correct it, then run Terraform again.
Error: release: already exists
```

dlorenc (approved):

PR #1316, opened 2023-09-03
Migrate kubectl image to track the `kubectl-default` virtual

mattmoor:


mattmoor:
I actually don't think we have the virtual packages we need to the kubernetes version streams.  Yes, `kubectl-latest` works here, but we lack something like it that installs particular versions as default for producing e.g. `kubectl:1.24`.  I am going to experiment with that here: https://github.com/wolfi-dev/os/pull/5099

mattmoor:
This should be RFAL!

imjasonh (approved):

PR #1321, opened 2023-09-05
Refactor buck2 to drop version tags

mattmoor:


mattmoor:
I'm going to batch these up

PR #1324, opened 2023-09-05
Refactor a number of images to drop version tags

mattmoor:
This refactors a large number of images to drop version tags and adopts the config module pattern.

I will leave this PR in draft mode while I continue to refactor more images, but I'm staging this to let CI validate things for me as I go 😁


images/dive/config/template.apko.yaml
```diff
@@ -1,6 +1,6 @@
 contents:
   packages:
-    - dive
```
imjasonh, line 3: There seem to be a bunch of these changes that are whitespace or ordering. Can you format or revert any of these, to avoid churn?

mattmoor: They also all coincide with renames, so it isn't zero churn.  I think these are largely from bringing back the packages manually :-/

mattmoor: tl;dr it wouldn't change the "153 files changed" :)

tcnghia (approved):

PR #1350, opened 2023-09-08
Have publisher optionally build dev variants w/ `build-dev = true`

mattmoor:


mattmoor:
I suspect that this'll save about 2400 more LoC in this repo based on some back-of-the-envelope calculations

imjasonh (approved):

PR #1357, opened 2023-09-11
Use tf-apko 0.10.4

imjasonh:
https://github.com/chainguard-dev/terraform-provider-apko/releases/tag/v0.10.4

### DIFF
--- a/tflib/publisher/main.tf
+++ b/tflib/publisher/main.tf
@@ -6,7 +6,7 @@ terraform {
     }
     apko = {
       source  = "chainguard-dev/apko"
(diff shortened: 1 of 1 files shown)


stormqueen1990 (approved):

tflib/publisher/main.tf
```diff
@@ -6,7 +6,7 @@ terraform {
     }
     apko = {
       source  = "chainguard-dev/apko"
-      version = "0.9.2"
```
mattmoor, line 9: You need to update the digest in the reproducibility check or this will fail postsubmit.

PR #1365, opened 2023-09-13
Add boring-registry image.

dlorenc:
I wrote a helm test for this but dropped it after. The upstream image has a helm chart but fails upon deployment without GCP or S3 credentials as well. The image is dead simple:

https://github.com/TierMobility/boring-registry/blob/main/Dockerfile

so I think the risk of this failing with GCS/S3 is low enough that it's not worth the complexity of adding a 3p service to our tests.

### Image Size

- [X] The Image is smaller in size than its common public counterpart.

Notes:

### Image Vulnerabilities

- [X] The Grype vulnerability scan returned 0 CVE(s).

Notes:

- [X] The container image was successfully loaded into a kind cluster.

Notes:

Notes: (see above, the service requires manual configuration against GCS/S3)

### Helm

- [X] A Helm chart has been provided and the container image is not working with the chart (please explain in the notes).

Notes: The image "works" with the helm chart, but by that I mean it fails in the same way as the upstream image.

### Processor Architectures

- [X] The image was built and tested for x86_64.
- [X] The image was built and tested for aarch64.

Notes:

- [X] Functional tests have been included and the tests are passing.  All tests have been documnted in the notes section.

Notes:

- [X] There has not been a request and/or there is no indication that this image needs tested on a public cloud provider.

Notes:

### Version

- [X] The package version is the latest version of the package.  The latest tag points to this version.

Notes:

### Dev Tag Availability

- [X] There is a dev tag available that includes a shell and apk tools (by depending on 'wolfi-base')

Notes:

- [X] The image runs as `nonroot` and GID/UID are set to 65532 or upstream default

### ENTRYPOINT

- [X] applications/servers/utilities set to call main program with no arguments e.g. [redis-server]

### CMD

- [X] For server applications give arguments to start in daemon mode (may be empty)

### Environment Variables

- [X] Environment variables not added and not required.

### SIGTERM

- [X] The image responds to SIGTERM (e.g., `docker kill $(docker run -d --rm cgr.dev/chainguard/nginx)`)

### Logs

- [X] Error logs write to stderr and normal logs to stdout. Logs DO NOT write to file.

### Documentation - README

- [X] A README file has been provided and it follows the README template.


stormqueen1990 (commented):
LGTM!

images/boring-registry/main.tf
```diff
@@ -0,0 +1,52 @@
...
+  depends_on = [module.test-latest]
+  digest_ref = module.latest.image_ref
+  tag        = "latest"
+}
+
+resource "oci_tag" "latest-dev" {
+  depends_on = [module.test-latest]
+  digest_ref = module.latest-dev.image_ref
+  tag        = "latest-dev"
+}
```
joshrwolf, line 52: any reason why we're not using the tagger module here?

joshrwolf:
```suggestion
module "tagger" {
  source = "../../tflib/tagger"

  depends_on = [module.test-latest]

  tags = merge(
    { for t in toset(concat(["latest"], module.version-tags.tag_list)) : t => module.latest.image_ref },
    { for t in toset(concat(["latest"], module.version-tags.tag_list)) : "${t}-dev" => module.latest-dev.image_ref },
  )
}
```

images/boring-registry/README.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+This will require specifying a storage provider and credentials to it using the appropriate flags.
+
+## Helm
+
+This image is a drop-in replacement for the upstream image, and can be used in their helm chart with an invocation like:
+
+```
+export HELM_EXPERIMENTAL_OCI=1
```
joshrwolf, line 112:
```suggestion
```

nit, since I realize this is just copy/pasta'd from upstream

images/boring-registry/main.tf
dlorenc: I just copied the helm image over which I guess uses the other pattern. Will try this one!

joshrwolf (approved):
lgtm, thanks!

PR #1366, opened 2023-09-13
add template directory to be used for new image configs

imjasonh:
cc @priyawadhwa @vaikas

### DIFF
--- a/images/TEMPLATE/README.md
+++ b/images/TEMPLATE/README.md
@@ -1,0 +1,11 @@
+This is a template for creating configs for new images.
+
+Include a brief description of the image here, and instructions about how to use it.
+
+When you're done, run `monopod readme` to update the top section of this file, and update the root README.md file.
+
+To get started, run:
+
+```sh
+cp -R images/TEMPLATE images/[new-image-name]
+```
--- a/images/TEMPLATE/README.md
+++ b/images/TEMPLATE/README.md
@@ -1,8 +1,6 @@
+# Chainguard Images Template
+
 This is a template for creating configs for new images.
-
-Include a brief description of the image here, and instructions about how to use it.
-
-When you're done, run `monopod readme` to update the top section of this file, and update the root README.md file.
 
 To get started, run:
 
@@ -10,6 +8,9 @@ To get started, run:
 cp -R images/TEMPLATE images/[new-image-name]
 ```
 
-Then, update your new `config/template.apko.yaml` file to specify packages you want the image to include, and any other necessary image config.
+1. Include a brief description of the image here, and instructions about how to use it.
+1. Run `monopod readme`, which will update the top section of this file and the root README.md file.
+1. Update your new `config/template.apko.yaml` file to specify packages you want the image to include, and any other necessary image config.
+1. Call this module from the `main.tf` in the root of this repo, in alphabetical order. (`lint.sh` will yell at you if you don't)
 
 If you need to support version streams, you can leave `packages` empty in `template.apko.yaml`, and instead add packages to the images using the `extra_packages` TF variable in `config/main.tf`.
--- a/images/TEMPLATE/README.md
+++ b/images/TEMPLATE/README.md
@@ -9,3 +9,7 @@ To get started, run:
 ```sh
 cp -R images/TEMPLATE images/[new-image-name]
 ```
+
+Then, update your new `config/template.apko.yaml` file to specify packages you want the image to include, and any other necessary image config.
+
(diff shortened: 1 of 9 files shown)


stormqueen1990 (commented):
Looks like this module also needs to be ignored in `lint.sh`

images/TEMPLATE/config/main.tf
```diff
@@ -0,0 +1,19 @@
+terraform {
+  required_providers {
+    apko = { source = "chainguard-dev/apko" }
+  }
+}
+
+variable "extra_packages" {
+  description = "The additional packages to install (e.g. crane)."
+  default     = ["crane"]
```
priyawadhwa, line 9: should this default be empty?

imjasonh: yep!

images/TEMPLATE/main.tf
```diff
@@ -0,0 +1,52 @@
...
+  depends_on = [module.test-latest]
+  digest_ref = module.latest.image_ref
+  tag        = "latest"
+}
+
+resource "oci_tag" "latest-dev" {
+  depends_on = [module.test-latest]
+  digest_ref = module.latest-dev.image_ref
+  tag        = "latest-dev"
+}
```
joshrwolf, line 52: do we want to use the tagger module here?

joshrwolf:
not to overcomplicate this, but considering how common it is for folders to contain multiple images, and how much delta there are between the two types of starting points, would this benefit from being 2 templates?

1. `TEMPLATE-single`: what this currently is
2. `TEMPLATE-multiple`: something akin to `images/vault`

or maybe:

```
images/
  TEMPLATE-SINGLE/
    this current PR
  TEMPLATE-MULTIPLE/
    a copy/pasta of something like images/vault
```

images/TEMPLATE/main.tf
imjasonh: Nope, the tagger module is moving out of the public repo, since we're only tagging `latest{-dev}` here.

imjasonh:
> my verbiage is up for debate, I don't think we formalize this anywhere yet

I love this idea, let's save it for a followup. I mostly want to get _something_ in before the onrush of new images coming this/next week 😄

images/TEMPLATE/config/template.apko.yaml
```diff
@@ -0,0 +1,19 @@
+contents:
```
vaikas, line 1: Can we add comments here on how to test if you have locally built packages? In the document, I had the example, of adding the keyrings, local package dir, etc.

At least the workflow I've been using for things that include many packages that need to be built, I keep on iterating on packages, and then test them locally.

Or maybe add that in the README.md file, or does it just magically work like this? 🤔

mattmoor: I think that's somewhat orthogonal to this change, but I think we should document it in `TERRAFORM.md` in the root of the repo.

images/TEMPLATE/tests/main.tf
```diff
@@ -0,0 +1,24 @@
...
+data "oci_exec_test" "version" {
+  digest = var.digest
+  script = "docker run --rm $IMAGE_NAME version"
+}
+
+// Invoke a script with the test.
+// TODO: Update or remove this test as appropriate.
+data "oci_exec_test" "manifest" {
+  digest = var.digest
+  script = "${path.module}/EXAMPLE_TEST.sh"
```
mattmoor, line 22: use `work_dir` instead?

images/TEMPLATE/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+To get started, run:
+
+```sh
+cp -R images/TEMPLATE images/[new-image-name]
+```
+
+Then, update your new `config/template.apko.yaml` file to specify packages you want the image to include, and any other necessary image config.
+
+If you need to support version streams, you can leave `packages` empty in `template.apko.yaml`, and instead add packages to the images using the `extra_packages` TF variable in `config/main.tf`.
```
mattmoor, line 15: This is missing adding the new module invocation to `main.tf` in the root of the repo (alphabetically!).

mattmoor:
@joshrwolf I like it, but I don't love the patterns we have for multi-component images yet, so let's see what we can do to improve those first?

joshrwolf:
> @joshrwolf I like it, but I don't love the patterns we have for multi-component images yet, so let's see what we can do to improve those first?

ack, which parts do you think need the most attention? happy to take a stab at improving them

mattmoor (approved):

PR #1396, opened 2023-09-16
Refactor `cert-manager` to eliminate boilerplate.

mattmoor:
This drops the YAML files entirely in favor of HCL-level templating.

This also factors `tflib/accts` to share the boilerplate around the accounts block.

cc @wlynch @imjasonh @joshrwolf


mattmoor:
I think it would be interesting trying to take a slightly more complex multi-component build like Tekton and migrate it to this style 🤔

imjasonh:
This change itself lgtm, and I support the direction with one caveat: we need better docs for how to use tf-apko from head to build and diff things, if this is going to be the paved/only path we use.

It's too painful today to make a change to go-apk, go mod replace it in apko, update and release tf-apko to see it in real usag e. If we can shorten that so folks (mainly Jon) can test things out in pure HCL then I think that's a win.

Also, if we stop using the apko CLI ourselves day to day, we should make that clear somehow in apko's docs and marketing. I don't want folks to find and use apko only to find out we're not hitting their bugs because we only use it via tf.

mattmoor:
> we need better docs for how to use tf-apko from head to build and diff things

We build a subset of this repo presubmit in tf-apko.  We can document the process for sure, but I’m pretty sure it’s standard tf provider development.

> It's too painful today to make a change to go-apk, go mod replace it in apko, update and release tf-apko to see it in real usag e. If we can shorten that so folks (mainly Jon) can test things out in pure HCL then I think that's a win.

I agree with the problem but disagree on the solution.  We need better tests in go-apk and apko, and frankly we always have.  If the only way to test things is to jump through all of these hoops then development is going to continue to suck.

These repos are already set up to build everything with both apko and tf-apko, so there is nothing really stopping us from adding a world build to apko like we have in tf-apko that tests apko still works too.  We just need a story for breaking digest changes if we are going to lean on that.  Frankly if that forces us to be more intentional with breaking changes, then I think that’s a win.

mattmoor:
[Here's the workflow](https://github.com/chainguard-dev/terraform-provider-apko/blob/main/.github/workflows/mega-module.yaml) I am talking about (mostly for others who are curious), but we'd need it to be `pull_request_target` in `apko` to reliably get the result I suggested above.

images/cert-manager/config/main.tf
```diff
@@ -0,0 +1,28 @@
...
+
+output "config" {
+  value = jsonencode({
+    contents = {
+      packages = concat(["cert-manager-${var.name}${var.suffix}"], var.extra_packages)
+    }
+    accounts = module.accts.block
+    entrypoint = {
+      command = "/usr/bin/${var.name}"
+    }
```
wlynch, line 26: I tried to do something like this originally with the apko_config, but ran into trouble since most of these fields are marked readonly.

Should we be dropping the read-only designation? (does this even matter w/ tf?)

images/cert-manager/config/main.tf
```diff
@@ -0,0 +1,28 @@
...
+variable "extra_packages" {
+  description = "Additional packages to install."
+  type        = list(string)
+  default     = []
+}
+
+module "accts" { source = "../../../tflib/accts" }
+
+output "config" {
+  value = jsonencode({
```
wlynch, line 19: Are we potentially double-encoding? is this a problem? 🤔 

```terraform
module "latest-dev" {
  config            = jsonencode(module.latest[each.key].config)
  ...
}
```

wlynch (approved):
lgtm, but a few questions

images/cert-manager/config/main.tf
mattmoor: This is a different config!

This line is the config that comes out of the non-dev version to ensure that everything is "locked" to the same packages in dev, except for anything new we add on top.

images/cert-manager/config/main.tf
mattmoor: I'm curious what you tried, but `apko_config` takes the string form and _returns_ the parsed/locked config, so it is output-only. However, what `apko_build` accepts is just this structure.

Hopefully that clarifies?

mattmoor:
I rebased this to fix the merge conflict.

imjasonh (approved):

PR #1409, opened 2023-09-19
bump release runner fanout to 4c runners

joshrwolf:
we're seeing:

```
  Warning  Evicted    3m28s  kubelet            The node had condition: [DiskPressure].
```

now on some legs of the build.

this bumps us up (slightly) to 4c runners, which also have _significantly_ larger disks

ref: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners#machine-sizes-for-larger-runners

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -52,7 +52,7 @@ jobs:
         modified-files: ${{ steps.build-filter.outputs.filter }}
 
   build-0:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

mattmoor:
There isn't a 4 core runner pool, so everything has been wedged since this landed.  I'm puzzled that we aren't seeing this in the private rebuilds?

mattmoor:
We applied this change downstream, but not here: https://github.com/chainguard-images/images/pull/1571

PR #1410, opened 2023-09-19
Adding missing GPG package for argoCD

sfc-gh-ddudek:
During running an argoCD image from chainguard there is a problem with running argocd-repo-server successfully.
The following error appears in the logs:
```
time="2023-09-19T10:16:46Z" level=fatal msg="exec: \"gpg\": executable file not found in $PATH"
```
Quick search-through the argo-cd repository, and yeah it uses gpg internally:
```
OUTPUT=$(gpg $ARGS 2>&1)
```
https://github.com/argoproj/argo-cd/blob/df714accc0184fea0f09cfa8a4e1d4ed4a6a7b4b/hack/gpg-wrapper.sh#L4

gpg is not in the packages list in the repo, hence adding it.

### DIFF
--- a/images/argocd/configs/latest.argocd.apko.yaml
+++ b/images/argocd/configs/latest.argocd.apko.yaml
@@ -3,6 +3,7 @@ contents:
     - busybox
     - argo-cd
     - argo-cd-compat
+    - gpg
 
 accounts:
   groups:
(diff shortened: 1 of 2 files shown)


joshrwolf:
thanks for the contribution!

just a heads up that this _should_ be pulled in as part of the `argocd-repo-server` package ([here](https://github.com/wolfi-dev/os/blob/cbc50ce131b2c7cceb80c70ddd1f5ba2ed9cd745/argo-cd-2.8.yaml#L75)).

I'll look into why that's not the case...

imjasonh (approved):

PR #1414, opened 2023-09-20
Add tests to Istio images

tcnghia:
- Test that the Helm charts are working
- Test that the sidecar injector / init container (iptable) are working
- Test that a simple VirtualService works

### DIFF
--- a/images/istio/tests/main.tf
+++ b/images/istio/tests/main.tf
@@ -14,6 +14,10 @@ variable "digests" {
   })
 }
 
+variable "namespace" {
+  description = "The namespace to install Istio in."
+}
+
 data "oci_exec_test" "proxy-version" {
   digest = var.digests.proxy
   script = "docker run --rm $IMAGE_NAME --version"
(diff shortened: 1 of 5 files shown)


images/istio/tests/main.tf
```diff
@@ -29,19 +33,126 @@ data "oci_exec_test" "operator-version" {
...
+  # there's no official helm chart for the istio operator
   repository = "https://stevehipwell.github.io/helm-charts/"
   chart      = "istio-operator"
 
   values = [jsonencode({
     image = {
-      repository = data.oci_string.ref.registry_repo
-      tag        = data.oci_string.ref.pseudo_tag
+      repository = data.oci_string.operator-ref.registry_repo
+      tag        = data.oci_string.operator-ref.pseudo_tag
```
mattmoor, line 50: Is it possible to point it at our images as well?

images/istio/tests/main.tf
```diff
@@ -29,19 +33,126 @@ data "oci_exec_test" "operator-version" {
...
+    # will be managed.
+    revision = local.namespace
+    pilot = {
+      image = var.digests.pilot
+    }
+    global = {
+      istioNamespace = local.namespace
+      # We have to trim the suffix and specify it in `image`, because this
+      # Helm chart does not like slashes in the image name.
+      hub = replace(data.oci_string.proxy-ref.registry_repo, "/istio-proxy", "")
```
mattmoor, line 92: 🤔 I'm puzzled at how this works

mattmoor (approved):

images/istio/tests/main.tf
tcnghia: This is our `operator` image. I think you mean pointing at our proxy and pilot image? That would be part of the IstioOperator resource that we will install. Will follow up with an additional test for that.

images/istio/tests/main.tf
tcnghia: If the proxy image is `gcr.io/test/istio/image/istio-proxy:unused@sha256:deadbeef`, the chart expects:
* `hub` = gcr.io/test/istio/image
* `image` = istio-proxy
* `tag` = `unused@...`

The `oci_string.foo.registry_repo` would give me gcr.io/test/istio/image/istio-proxy, hence I need to chop that last part off.

PR #1421, opened 2023-09-21
bump apko publisher module and cosign provider

imjasonh:
Supercedes https://github.com/chainguard-images/images/pull/1420 which fails because it only updates the publisher.


k4leung4:
this should be safe to close as there are no changes

PR #1423, opened 2023-09-21
feat: add scaffold command to monopod

stormqueen1990:
### Changes

* Create a new `scaffold` command in `monopod` to aid creation of new images.
* Reorganize the `images/TEMPLATE` folder to a flat structure to make it easier to read the template files.
* Change templates to work with Go `text/templates`.


images/TEMPLATE/main-config.tf
```diff
@@ -12,7 +12,7 @@ variable "extra_packages" {
 }
```
pdeslaur, line 12: Should `TODO YOUR PACKAGE HERE` be replaced with `.PackageName`? If so, we could also update the `apko` YAML to include the comment:

`# {{ .PackageName }} comes in via var.extra_packages`

priyawadhwa (commented):
This LGTM! Maybe we could add a test to generate a sample scaffold & compare to the expected output.

vaikas (commented):
Thanks for doing this!

images/TEMPLATE/main-config.tf
stormqueen1990: Yes that makes sense to me!

stormqueen1990: Done!

stormqueen1990:
> This LGTM! Maybe we could add a test to generate a sample scaffold & compare to the expected output.

Yes! That's a great idea, I'll add it.

imjasonh (approved):

PR #1424, opened 2023-09-21
crossplane tests: enable pushing to a private repo

imjasonh:
These tests pass when pushing to cgr.dev/chainguard, where anybody can pull, but fail when pushed to another non-public repo on cgr.dev (or, I suppose, any private repo anywhere).

The tests fail because the Crossplane control plane running on the cluster doesn't have permission to pull the image, which it needs to do to extract provider metadata from it.

This PR updates the test to check if the cgr cred helper is installed, and if so, populate an ephemeral Docker config, and populate a Kubernetes Secret with those creds, and tell the Crossplane control plane to use the secret to pull.

This passed for me locally with:

```
terraform apply -var=target_repository=cgr.dev/imjasonh.dev -target=module.crossplane-aws -auto-approve
```

### DIFF
--- a/images/crossplane-aws/tests/install.sh
+++ b/images/crossplane-aws/tests/install.sh
@@ -1,6 +1,27 @@
 #!/usr/bin/env bash
 
-set -o errexit -o nounset -o errtrace -o pipefail -x
+set -o errexit -o nounset -o errtrace -o pipefail
+
+# If the image is on cgr.dev and we have a cred helper for that available, use
+# it to generate creds we can pass to the Crossplane control plane.
+# This is necessary when pushing to a private repo, since the Crossplane
+# control plane needs to pull the image to get configuration from it.
+tmp=$(mktemp -d)
+echo "{}" > ${tmp}/config.json
+if [ -x "$(command -v docker-credential-cgr)" ]; then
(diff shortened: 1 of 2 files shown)


images/crossplane-aws/tests/install.sh
```diff
@@ -1,6 +1,27 @@
...
+set -o errexit -o nounset -o errtrace -o pipefail
+
+# If the image is on cgr.dev and we have a cred helper for that available, use
+# it to generate creds we can pass to the Crossplane control plane.
+# This is necessary when pushing to a private repo, since the Crossplane
+# control plane needs to pull the image to get configuration from it.
+tmp=$(mktemp -d)
+echo "{}" > ${tmp}/config.json
+if [ -x "$(command -v docker-credential-cgr)" ]; then
+  token=$(echo "cgr.dev" | docker-credential-cgr get)
```
mattmoor, line 12: Can you pass in the registry via an environment variable extracted from the `IMAGE_NAME`?

mattmoor (approved):

images/crossplane-aws/tests/install.sh
imjasonh: I thought about it, I can do that in a followup when/if it's a problem.

PR #1425, opened 2023-09-21
aws-efs-csi-driver: Fix /csi permissions

jonjohnsonjr:
When this boots up, it attempts to delete `/csi/csi.sock` and then listen on it, but does not have permissions to do so.


denhamparry (approved):
LGTM!

ajayk (approved):

PR #1435, opened 2023-09-22
add structure test check that /etc/os-release is Wolfi

imjasonh:
This resolves a TODO and ensures that no non-Wolfi images sneak through.

Some repos (static, busybox, git) can include some Alpine-based images, so for now we'll just ignore those and check the normal repos.

### DIFF
--- a/tflib/publisher/main.tf
+++ b/tflib/publisher/main.tf
@@ -60,6 +60,11 @@ data "oci_exec_test" "check-reproducibility" {
   timeout_seconds = 600
 }
 
+locals {
+  parts           = split("/", var.target_repository)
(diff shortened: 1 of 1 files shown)


tflib/publisher/main.tf
```diff
@@ -75,7 +80,12 @@ data "oci_structure_test" "structure" {
     //}
     files { path = "/etc/ssl/certs/ca-certificates.crt" }
     files { path = "/lib/apk/db/installed" }
-    files { path = "/etc/os-release" } // TODO: we can check the contents
+    files {
+      path = "/etc/os-release"
+      // Images should report as "Wolfi".
+      // Some repos are special and some report as Alpine.
+      regex = local.is_maybe_alpine ? "" : "PRETTY_NAME=\"Wolfi\""
```
mattmoor, line 87: Nack

PR #1436, opened 2023-09-22
use the already installed helm chart for the coredns smoke tests

joshrwolf:
simplifies the flakey bash/docker tests by just using the already deployed helm chart and a `kubernetes_job` that employs more stable, builtin retrys.

the same test is performed (a `TXT` query on a custom `Corefile`), except this time within the cluster.


images/coredns/tests/main.tf
```diff
@@ -13,9 +13,43 @@ data "oci_string" "ref" {
   input = var.digest
 }
 
-data "oci_exec_test" "nslookup" {
-  digest = var.digest
-  script = "${path.module}/nslookup-with-Corefile.sh"
+locals {
+  # NOTE: The coredns helm values are a pain to translate into terraform, so just cheat and define them as inline yaml
```
mattmoor, line 17: you could put it in a file and use `file()`?

mattmoor (approved):

mattmoor (approved):

joshrwolf:
reminder to myself or the next person that this helm chart defaults to a cluster wide service `isClusterService`, which translates to a few hardcoded helm resources.

for the next iteration of these tests, we should disable that and make sure the functionality is still ~equivalent, so we can have multiple of theses test modules in a single cluster.

PR #1445, opened 2023-09-23
cut out a bunch of unused stuff from monopod 

imjasonh:
- remove `image.yaml`s from everywhere, and code to handle it
  - 🚨 this removes where we signal a stable image vs an experimental one -- if we still want this, we can signal it some other way
  - this file used to be used to configure variants, extra tags, etc., all of which are now in TF
- remove globals.yaml, it's unused, and code to handle it
- remove GH workflow infra to push to ghcr.io

Helfpul links:
- [the root README.md](https://github.com/imjasonh/images/blob/no-image-yaml/README.md)
- [an example image README.md](https://github.com/imjasonh/images/blob/no-image-yaml/images/bash/README.md)
- [`monopod` code changes](https://github.com/chainguard-images/images/pull/1445/files?file-filters%5B%5D=.go&show-deleted-files=true&show-viewed-files=true)


.github/actions/release-image-terraform/action.yml
```diff
@@ -84,27 +76,25 @@ runs:
...
+          export TF_VAR_target_repository=cgr.dev/chainguard
           echo ::group::terraform init
           terraform init
           echo ::endgroup::
           terraform apply -auto-approve "-target=module.${MODULE}" -json | tee /tmp/${MODULE}.tf.json | jq -r '.["@message"]'
         else
-          cd ${{ inputs.terraformDirectory }}
-          export TF_VAR_target_repository=${{ inputs.apkoBaseTag }}
+          cd ${MODULE}
+          export TF_VAR_target_repository=cgr.dev/chainguard/${{ inputs.imageName }}
```
imjasonh, line 97: I think we'll want this to be an input.

monopod/pkg/images/images.go
```diff
@@ -1,82 +1,20 @@
...
-type ImageManifestVariantApkoSubvariant struct {
-	Suffix  string   `yaml:"suffix"`
-	Options []string `yaml:"options"`
-}
-
-type ImageManifestVariantApkoExtractTagsFrom struct {
-	Package string   `yaml:"package"`
-	Prefix  string   `yaml:"prefix"`
-	Exclude []string `yaml:"exclude"`
+	ImageName string `json:"imageName"`
```
imjasonh, line 10: If there's nothing else in here we want to keep, this can just be a `string`. We usually consume it as a `[]Image`, which could be a `[]string`.

.github/actions/release-image-terraform/action.yml
```diff
@@ -84,27 +76,25 @@ runs:
...
 
-        MODULE=$(basename ${{ inputs.terraformDirectory }})
+        MODULE=${{ inputs.imageName }}
         echo "module=${MODULE}" >> $GITHUB_OUTPUT
 
         # When we have a "mega module", use it to invoke the module
         # for the appropriate image.
         if [ -f "main.tf" ]; then
-          export TF_VAR_target_repository=$(dirname ${{ inputs.apkoBaseTag }})
+          export TF_VAR_target_repository=cgr.dev/chainguard
```
mattmoor, line 90: This may break downstream rebuilds, we should think about that.

.github/actions/release-image-terraform/action.yml
```diff
@@ -84,27 +76,25 @@ runs:
...
         set -x -o pipefail
         env | grep '^TF_VAR_'
 
-        MODULE=$(basename ${{ inputs.terraformDirectory }})
+        MODULE=${{ inputs.imageName }}
         echo "module=${MODULE}" >> $GITHUB_OUTPUT
 
         # When we have a "mega module", use it to invoke the module
         # for the appropriate image.
         if [ -f "main.tf" ]; then
```
mattmoor, line 89: I think this is always true now.  This was migration code.

images/bash/config/main.tf
```diff
@@ -0,0 +1,20 @@
...
+}
+
+output "config" {
+  value = jsonencode({
+    contents = {
+      packages = concat([
+        "bash",
+        "busybox",
+        "curl",
+      ], var.extra_packages)
```
mattmoor, line 14: did you conflate these two changes?

monopod/pkg/commands/readme.go
```diff
@@ -259,62 +209,18 @@ func getRootReadmeContents(allImages []images.Image, badgeRootUrl string, defaul
 	// TODO: Remove in October
```
mattmoor, line 209: We should remember to do this.

mattmoor (commented):
Generally LGTM, let's get it in?

.github/actions/release-image-terraform/action.yml
imjasonh: Yep, this should be an input to the action.

images/bash/config/main.tf
imjasonh: I based this change on the other one (#1444) on accident, but it was nice to see it trigger bash's leg.

mattmoor (approved):

PR #1451, opened 2023-09-24
Refactor `cert-manager` to virtual/latest pattern

mattmoor:
We have 1.13, but it looks like it's missing the provides, so we'll need to go back and add that.

https://github.com/wolfi-dev/os/blob/main/cert-manager-1.13.yaml

cc @k4leung4 @wlynch


mattmoor:
Ok, once https://github.com/wolfi-dev/os/pull/5901 lands, this will automatically float `:latest` to 1.13

k4leung4 (approved):

tcnghia (approved):

PR #1455, opened 2023-09-25
Refactor the remaining dev variants.

mattmoor:
Now the only reference to the `dev` module is from the publisher :tada:

### DIFF
--- a/images/git/alpine.tf
+++ b/images/git/alpine.tf
@@ -2,51 +2,32 @@ module "latest-alpine" {
   providers = {
     apko = apko.alpine
   }
-  for_each = local.accounts
-  source   = "../../tflib/publisher"
-
-  name = basename(path.module)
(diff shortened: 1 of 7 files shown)


images/git/alpine.tf
```diff
@@ -2,51 +2,32 @@ module "latest-alpine" {
...
-  providers = {
-    apko = apko.alpine
-  }
-  for_each = local.accounts
-  source   = "../../tflib/publisher"
-
-  name = basename(path.module)
-
-  target_repository = var.target_repository
-  config            = file("${path.module}/configs/latest.alpine.${each.key}.apko.yaml")
```
mattmoor, line 26: This was a bug, it could have led to inconsistent dev versions :(

tcnghia (approved):

PR #1480, opened 2023-09-28
Update redis image so we can invoke it from images-private

mamccorm:
## Summary
Few small tweaks to the redis image to bring it (closer) in line with our newer images. Namely this will allow us to invoke it from images-private and pass in different redis package versions, as well as fips openssl config module.

Being carried out in prep for: [image-requests/issues/365](https://github.com/chainguard-dev/image-requests/issues/365)

## Testing
Checked I could build and run the image locally. Note i'm on macOS and the included test failed so i needed to disable it when doing local runs. Expecting its some mac issue and going to check the test passes in the CI run. Regardless I didn't make any changes to the existing test.

### DIFF
--- a/images/redis/configs/latest.apko.yaml
+++ b/images/redis/configs/latest.apko.yaml
@@ -1,7 +1,6 @@
 contents:
   packages:
-    - redis
-    - redis-cli
+    # redis and redis-cli come in via var.extra_packages
     - busybox
 
 accounts:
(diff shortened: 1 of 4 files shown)


mattmoor (approved):
This looks good.  Generally when I've been doing these I've also been dropping the `version-tags` module and adding explicit version stream builds to `images-private` for all of the versions we care about, so if that's next for `redis`, then I'd recommend doing that here as well.

mamccorm:
Thanks for the review @mattmoor, makes sense. i'm chunking up what i'm refactoring as i'm still finding my way, but that makes sense. i'm going to do what you described for the fips image variant and depending how I get on i might re-factor the existing redis private image also

imjasonh (approved):

PR #1489, opened 2023-09-29
Fluentd module path

found-it:
Use a module path environment variable to dynamically set where the splunk tests are run


images/fluentd/tests/main.tf
```diff
@@ -11,4 +11,8 @@ variable "digest" {
 data "oci_exec_test" "test" {
   digest = var.digest
   script = "${path.module}/01-test.sh"
+  env {
+    name  = "MODULE_PATH"
+    value = path.module
+  }
```
mattmoor, line 17: I think you want `working_dir = path.module` instead, then `pwd` in the scripts will be more predictable

mattmoor: `pwd` will be the `path.module` 😁

found-it: :chefs-kiss"

mattmoor (approved):

PR #1495, opened 2023-09-30
Remove usr/local from entrypoint path.

dlorenc:
This was flagging our linters. The upstream images *do* place the binaries in usr/local, but the entrypoint doesn't contain that path: https://github.com/crossplane-contrib/provider-aws/blob/master/cluster/images/provider-aws/Dockerfile


dlorenc:
This needs https://github.com/wolfi-dev/os/pull/6060 and https://github.com/wolfi-dev/os/pull/6057

stormqueen1990 (approved):

PR #1529, opened 2023-10-03
enable version-tagging crossplane-{aws,azure} images

imjasonh:
This enables version-tagging crossplane provider images, in addition to tagging `:latest`.

### DIFF
--- a/images/crossplane-aws/main.tf
+++ b/images/crossplane-aws/main.tf
@@ -47,9 +47,21 @@ module "test-latest" {
   digests = { for k, v in module.latest : k => v.image_ref }
 }
 
-resource "oci_tag" "latest" {
-  for_each   = toset(local.components)
+module "version-tags" {
+  for_each = toset(local.components)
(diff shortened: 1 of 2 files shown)


images/crossplane-aws/main.tf
```diff
@@ -47,9 +47,21 @@ module "test-latest" {
...
 }
 
-resource "oci_tag" "latest" {
-  for_each   = toset(local.components)
+module "version-tags" {
+  for_each = toset(local.components)
+  source   = "../../tflib/version-tags"
+  package  = "crossplane-provider-aws-${each.key}"
+  config   = module.latest[each.key].config
+}
```
mattmoor, line 55: I think this should live downstream.  I don't think we should be adding more `version-tags` in `images`

imjasonh: I'm fine with that. However a customer is waiting on this and I'd like to ask to clean this up later, with the rest.

mattmoor (approved):

PR #1535, opened 2023-10-04
istio images tests: use more robust splitting logic

tcnghia:
The current splitting logic assumes the last part of the repository is named `istio-proxy` or `istio-install-cni`. This removes such assumption so this test can be reused better.

### DIFF
--- a/images/istio/tests/main.tf
+++ b/images/istio/tests/main.tf
@@ -100,12 +100,12 @@ resource "helm_release" "istiod" {
       # If the registry_repo is gcr.io/my/repo/istio-proxy, we need to set
       #   hub   = gcr.io/my/repo
       #   image = istio-proxy
-      hub = replace(data.oci_string.proxy-ref.registry_repo, "//[^/]*$/", "")
+      hub = dirname(data.oci_string.proxy-ref.registry_repo)
       proxy = {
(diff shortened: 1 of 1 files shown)


images/istio/tests/main.tf
```diff
@@ -96,16 +95,19 @@ resource "helm_release" "istiod" {
...
-      tag = data.oci_string.proxy-ref.pseudo_tag
+      # These Helm charts do not like slashes in the image name.
+      #
+      # If the registry_repo is gcr.io/my/repo/istio-proxy, we need to set
+      #   hub   = gcr.io/my/repo
+      #   image = istio-proxy
+      hub = replace(data.oci_string.proxy-ref.registry_repo, "//[^/]*$/", "")
       proxy = {
-        image = "istio-proxy"
+        image = replace(data.oci_string.proxy-ref.registry_repo, "/^.*//", "")
```
mattmoor, line 105: This pattern seems wrong 🤔

tcnghia: changed to use `basename` / `dirname` now.

tcnghia: (the pattern is weird because `replace` only treat the search string as regexp if it begins and ends w/ slashes -- very confusing)

mattmoor (approved):

PR #1537, opened 2023-10-04
New image: smarter-device-manager

mamccorm:
## Summary
New image for smarter-device-manager.  Docker and helm tests added.

### Image Size

- [x] The Image is smaller in size than its common public counterpart.
- [] The Image is larger in size than its common public counterpart (please explain in the notes).

### Image Vulnerabilities

- [x] The Grype vulnerability scan returned 0 CVE(s).
- [] The Grype vulnerability scan returned > 0 CVE(s) (please explain in the notes).

Notes:

- [x] The container image was successfully loaded into a kind cluster.
- [ ] The container image could not be loaded into a kind cluster (please explain in the notes).

Notes:

- [x] The application is accessible to the user/cluster/etc. after start-up.
- [ ] The application is not accessible to the user/cluster/etc. after start-up. (please explain in the notes).

Notes:

### Helm

- [x] A Helm chart has been provided and the container image can be used with the chart.  If needed, please add a -compat package to close any gaps with the public helm chart.
- [ ] A Helm chart has been provided and the container image is not working with the chart (please explain in the notes).
- [ ] A Helm chart was not provided.

Notes:

### Processor Architectures

- [x] The image was built and tested for x86_64.
- [ ] The image could not be built for x86_64 (please explain in the notes).
- [x] The image was built and tested for aarch64.
- [ ] The image could not be built for aarch64. (please explain in the notes).

Notes:

- [x] Functional tests have been included and the tests are passing.  All tests have been documnted in the notes section.

Notes:

- [x] There has not been a request and/or there is no indication that this image needs tested on a public cloud provider.
- [ ] The container image has been tested successfully on a public cloud provider (AWS, GCP, Azure).
- [ ] The container image has not been tested successfully on a public cloud provider (AWS, GCP, Azure) (please explain in the notes).

Notes:

### Version

- [x] The package version is the latest version of the package.  The latest tag points to this version.
- [ ] The package version is the not the latest version of the package (please explain in the notes).

Notes:

### Dev Tag Availability

- [ ] There is not a dev tag available that includes a shell and apk tools (by depending on 'wolfi-base') (please explain in the notes).

Notes:

- [x] The image requires a non-standard username or non-standard GID/UID (please explain in the notes).

Notes:
The image must be run as root in order to be able to interact with volumes and resources within the cluster. Attempting to switch the image to run as non-root will result in errors such as: 'Could not start device plugin: listen unix /var/lib/kubelet/device-plugins/smarter-ttyS0.sock: bind: permission denied'.

### ENTRYPOINT

- [X] applications/servers/utilities set to call main program with no arguments e.g. [redis-server]
- [ ] applications/servers/utilities not set to call main program with no arguments e.g. [redis-server] (please explain in the notes)
- [ ] base images leave empty.
- [ ] base image and not empty (please explain in the notes).
- [ ] dev variants is set to entrypoint script that falls back to system.
- [ ] dev variants is not set to entrypoint script that falls back to system (please explain in the notes).

### CMD

- [X] For server applications give arguments to start in daemon mode (may be empty)

### Environment Variables

- [X] Environment variables added.
- [] Environment variables not added and not required.

We have to override the default 'config' path. This is handled as part of the entrypoint in the image.

### SIGTERM

- [X] The image responds to SIGTERM (e.g., `docker kill $(docker run -d --rm cgr.dev/chainguard/nginx)`)

### Logs

- [X] Error logs write to stderr and normal logs to stdout. Logs DO NOT write to file.

### Documentation - README

- [X] A README file has been provided and it follows the README template.


images/smarter-device-manager/config/latest.apko.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - username: nonroot
+      uid: 65532
+      gid: 65532
+  # smarter-device-manager requires root to interact with directories creaed by device plugins
+  run-as: 0
+
+entrypoint:
+  command: /usr/bin/smarter-device-management
+
+cmd: -config=/etc/smarter-device-manager/conf.yaml -logtostderr=true -v=2
```
rawlingsj, line 19: nit - I wonder if the log level can be made configureable?  I.e. via an env var or mounted config?  Just thinking that folks may want the ability to change this for production or when debugging?

mamccorm: Sure I can take a look. with this particular application though it'll simply exit and log nothing if there is an issue, i.e:

```
# smarter-device-management -config=/etc/smarter-device-manager/conf.yaml
#
```

Nothing informational either so not the easiest to debug. Kind of my thinking for enabling this and it didn't look to verbose. That said I can certainly look to put behind a flag and document if preferred

rawlingsj: ok well let's not block on this, see what you can find.

rawlingsj (approved):

images/smarter-device-manager/config/latest.apko.yaml
mamccorm: Also meant to add thats what the vendor currently does in their image, so at least we're consistent in a sense:
- https://gitlab.com/arm-research/smarter/smarter-device-manager/-/blob/master/Dockerfile#L22

PR #1562, opened 2023-10-05
This corrects the suffix placement for several images.

mattmoor:
We use the pattern `${{package.name}}-foo` so the suffix comes after the parent package name, but prior to the subcomponent.


imjasonh (approved):

mattmoor:
Blah, I missed one place in the `rekor` config :-/

imjasonh (approved):

PR #1570, opened 2023-10-07
Revert "bump release runner fanout to 4c runners"

mattmoor:
Reverts chainguard-images/images#1409

PR #1571, opened 2023-10-07
This was applied downstream but not here.

mattmoor:
ref: https://github.com/chainguard-images/images/pull/1570


mattmoor:
I expect two failures: `bank-vaults` and `metacontroller` which are consistently failing on a Helm 403 that needs investigation.

The leg to watch here is `kubeflow-katib`, which has been failing due to disk pressure, which this change should resolve.

imjasonh (approved):

mattmoor:
The two failures are OCI Helm tests and failing at HEAD, so bypassing those presubmits to land the disk change and unblock `kubeflow-katib`

PR #1575, opened 2023-10-07
Drop the extra checkout.

mattmoor:
This action used to be used to build stuff for other repos, and this would check out the code to build things.  Now this action is only used by chainguard-images/images and that already needs to be checked out to load the action definition itself.  So we should be able to safely drop these steps now.


k4leung4 (approved):

mattmoor:
I think we're just going to kill these instead.  h/t @imjasonh

PR #1577, opened 2023-10-08
envoy: update template to use latest tflibs

tcnghia:


imjasonh (approved):

images/envoy/main.tf
```diff
@@ -1,33 +1,31 @@
...
+
 variable "target_repository" {
   description = "The docker repo into which the image and attestations should be published."
 }
 
-module "latest" {
-  source = "../../tflib/publisher"
-
-  name = basename(path.module)
+module "config" { source = "./configs" }
```
mattmoor, line 11: super nit: I've typically been renaming these `config` because the plural seems misleading.

PR #1580, opened 2023-10-08
remove {build,release}-image-terraform actions

imjasonh:
After recent refactors, the only place these actions are used are inside this repo.

They've both been reduced to doing effectively the same thing:

1. install tools like `terraform`, `cosign`, `crane`
2. run `terraform init && terraform apply`
3. surface warnings/errors reported from (2) as JSON

In `release.yaml` we set up creds to push to cgr.dev. In `presubmit-build.yaml` we push to localhost:5000.

We'll find out if I missed anything in the presubmit leg if the presubmit passes. We won't be able to test the release workflow without doing it postsubmit.


mattmoor:
We need to inline it here first: https://github.com/chainguard-images/images-private/blob/6c2b4a1cccf740544e311ded0f0617c9522726f3/.github/workflows/release.yaml#L229

imjasonh:
This should BRFAL.

mattmoor (approved):
LGTM, but let's wait until tomorrow to land this so we can iterate if/when things break.

mattmoor (approved):

PR #1593, opened 2023-10-11
use bitnami helm chart to test prometheus-node-exporter

imjasonh:


joshrwolf:
can we get rid of the oci_exec tests for this one? 

the new helm charts readiness sufficiently replaces it imo

joshrwolf (approved):

PR #1595, opened 2023-10-11
add redis-bitnami

joshrwolf:
Adds `redis-bitnami`, `redis-bitnami-sentinel`, and `redis-bitnami-cluster`.

Noteworthy:

- `redis-bitnami-cluster` appears unused in the helm chart, but I'm including it for completeness
- our existing `redis-sentinel` remains untouched. I'm chalking this up to "legacy" and leaving it as is to avoid affecting any existing customers using it. moving forward, we want to replace that with these images.
  - `redis-sentinel` is a simple symlink to `redis-server` that the official `redis` distribution doesn't isolate, but bitnami chooses to use different startup scripts for this, hence the additional image

depends on: https://github.com/wolfi-dev/os/pull/6593

### DIFF
--- a/images/redis-bitnami/README.md
+++ b/images/redis-bitnami/README.md
@@ -1,0 +1,36 @@
+<!--monopod:start-->
+# redis (bitnami)
+| | |
+| - | - |
+| **OCI Reference** | `cgr.dev/chainguard/redis-bitnami` |
+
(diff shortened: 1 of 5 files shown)


joshrwolf:
closing in favor of #1617

PR #1607, opened 2023-10-11
Try deploying our Postgres with Bitnami Helm chart.

mattmoor:


images/postgres/tests/main.tf
```diff
@@ -8,6 +8,8 @@ variable "digest" {
   description = "The image digest to run tests over."
 }
 
+data "oci_string" "ref" { input = var.digest }
+
 data "oci_exec_test" "version" {
```
imjasonh, line 13: (not a binding review comment) These version tests don't really add much now that we're testing actual stuff.

imjasonh (approved):

images/postgres/tests/main.tf
mattmoor: Agreed, maybe we should do a pass to clean them up everywhere we feel we're doing better?  Same for `--help` tests?

PR #1615, opened 2023-10-12
Add monopod readme commands and templates

jamonation:
So here's a big PR for readmes. It will need some careful review and socializing, since it will make monopod yell and fail in CI if a readme is not well structured. 

First, new commands are:

1. `monopod readme new --image imageName`
2. `monopod readme render --image imageName`

Respectively, these will:
1. Create an `images/imageName/metadata.hcl` file for an existing image
2. Render an `images/imageName/metadata.hcl` file into `images/imageName/README.md`

Changed commands include:

1. `monopod scaffold package-name ...`: this does the usual scaffolding, including a `README.md` with a placeholder set of `<!--body:start--><!--body:end-->` tokens. It also creates a `metadata.hcl` file with various fields to fill out. The `short_description` field is mandatory.
2. `monopod readme --check`: this will validate all metadata.hcl and README.md files are structurally sound and fields are populated.
3.  `monopod readme`: this rerenders all READMEs. It is equivalent to running `monopod readme render --image ...` in a for loop across all image modules.

I've included a test in `readme_test.go` to validate that readme rendering works. I also tested in `images-private` and it doesn't break anything there or behave any differently.


jamonation:
Closing in favour of #1888 and #1889

PR #1624, opened 2023-10-13
Drop version tags for vault and vault-k8s

Dentrax:
Drop version tags for `vault` and `vault-k8s` images.

### DIFF
--- a/images/vault/main.tf
+++ b/images/vault/main.tf
@@ -1,7 +1,6 @@
 terraform {
   required_providers {
-    apko = { source = "chainguard-dev/apko" }
-    oci  = { source = "chainguard-dev/oci" }
+    oci = { source = "chainguard-dev/oci" }
   }
 }
 
--- a/images/vault/main.tf
+++ b/images/vault/main.tf
@@ -1,17 +1,7 @@
-locals {
-  components = toset([
-    "vault",
-    "vault-k8s",
-  ])
-
-  packages = {
-    vault     = "vault-1.14",
-    vault-k8s = "vault-k8s",
-  }
-
-  repositories = {
-    vault     = var.target_repository,
-    vault-k8s = "${var.target_repository}-k8s",
+terraform {
+  required_providers {
+    apko = { source = "chainguard-dev/apko" }
+    oci  = { source = "chainguard-dev/oci" }
   }
 }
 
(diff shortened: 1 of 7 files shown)


imjasonh (approved):

images/vault-k8s/config/template.apko.yaml
```diff
@@ -1,14 +1,12 @@
 contents:
   packages:
-    - vault-k8s
-    - libcap
 
 accounts:
   groups:
-    - groupname: vault
+    - groupname: vault-k8s
```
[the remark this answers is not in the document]
Dentrax: Actually we don't. But it'd be great to make those consistent with the package name, as we did follow similar way for other packages.

Maybe we should refactor it for all the images and just set: `nonroot`. 🤷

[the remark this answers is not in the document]
imjasonh: I'd like to make as few changes as possible in here, and consider renaming the user as a separate change. Maybe it should just be named `nonroot` and we should have better processes/linters in place to encourage that standard.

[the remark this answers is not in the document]
Dentrax: > Maybe it should just be named nonroot and we should have better processes/linters in place to encourage that standard.

+1, can i file a proposal for this? to enforce `nonroot` everywhere. if accepted, then i can create a linter rule for this.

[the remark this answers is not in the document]
Dentrax: Reverted back to `vault`

images/vault/tests/main.tf
```diff
@@ -4,18 +4,11 @@ terraform {
...
 }
 
-variable "digests" {
-  description = "The image digests to run tests over."
-  type = object({
-    vault     = string
-    vault-k8s = string
-  })
+variable "digest" {
+  description = "The image digest to run tests over."
```
mattmoor, line 8: This is a breaking change for us downstream.  We discussed on slack keeping this as a single module, let's do that unless we have a really good reason not to.

mattmoor (changes requested):

images/vault/tests/main.tf
Dentrax: Done!

images/vault/vault.tf
```diff
@@ -0,0 +1,9 @@
+module "vault-config" { source = "./configs/vault" }
+
+module "vault" {
+  source            = "../../tflib/publisher"
+  name              = basename(path.module)
+  target_repository = var.target_repository
```
Dentrax, line 6: is it correct? or should we have `-vault` suffix here?

images/vault/main.tf
```diff
@@ -1,53 +1,38 @@
...
-    vault     = "vault-1.14",
-    vault-k8s = "vault-k8s",
-  }
-
-  repositories = {
-    vault     = var.target_repository,
-    vault-k8s = "${var.target_repository}-k8s",
+terraform {
+  required_providers {
+    apko = { source = "chainguard-dev/apko" }
```
mattmoor, line 3: you don't need this provider

images/vault/vault.tf
mattmoor: This is perfect otherwise it is `vault-vault`

images/vault/main.tf
```diff
@@ -1,53 +1,38 @@
...
+    oci  = { source = "chainguard-dev/oci" }
```
mattmoor, line 4:
```suggestion
    oci = { source = "chainguard-dev/oci" }
```

mattmoor (approved):

PR #1625, opened 2023-10-13
Output the fluentd splunk variant config

found-it:
Just moving the splunk variant config to an output of the fluentd config module so it can be consumed for the version tagging


found-it:
After speaking with @mattmoor this is no longer necessary

PR #1627, opened 2023-10-14
Withdraw unused external-attacher image

tcnghia:


pdeslaur:
We should be able to drop this PR in favor of #1633

PR #1648, opened 2023-10-16
Allow `withdrawn-*.txt` files to presubmit build CI

pdeslaur:
This change to `monopod` should prevent the CI matrix builder from invoking a build for every image in the repo when the `withdrawn-images.txt` and `withdrawn-repos.txt` files are updated.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	golang.org/x/exp v0.0.0-20230811145659-89c5cff77bcb
 	golang.org/x/text v0.13.0
 	gopkg.in/yaml.v3 v3.0.1
+	k8s.io/apimachinery v0.27.3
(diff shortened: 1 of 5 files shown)


jonjohnsonjr (approved):

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	golang.org/x/exp v0.0.0-20230811145659-89c5cff77bcb
 	golang.org/x/text v0.13.0
 	gopkg.in/yaml.v3 v3.0.1
+	k8s.io/apimachinery v0.27.3
```
jonjohnsonjr, line 14: I don't love this.

pdeslaur: That's fair. I import it to use the set stuff, but it might be better to just do a map[string]bool like before.

PR #1657, opened 2023-10-17
cilium: use an on-demand test cluster

tcnghia:
cilium needs a cluster without CNI, so we can't use the current CI cluster.

This PR creates another cluster in a separate context to test, and then tears down.


cpanato (approved):

images/cilium/tests/main.tf
```diff
@@ -16,36 +24,38 @@ data "oci_exec_test" "operator-version" {
...
+  depends_on = [data.oci_exec_test.create-cluster]
+  repository = "https://helm.cilium.io"
+  chart      = "cilium"
+  name       = "cilium"
+  namespace  = "kube-system"
+
+  values = [jsonencode({
+    operator = {
+      image = {
+        override = var.digests.operator
```
joshrwolf, line 42: do we need the agent image as well? or is that coming later?

images/cilium/tests/main.tf
```diff
@@ -16,36 +24,38 @@ data "oci_exec_test" "operator-version" {
...
+
+data "oci_exec_test" "wait-for-pods" {
+  depends_on = [helm_release.cilium]
+  # Due to lack of CNI, all pods in `kube-system` will be in `Pending` state.
+  # If they are 'Ready', then our installation is working.
+  script = "kubectl --context k3d-cilium-test wait --for condition=Ready -n kube-system pods --all --timeout=300s"
+  digest = var.digests.operator
+}
+
+data "oci_exec_test" "delete-cluster" {
```
joshrwolf, line 57: does this delete the cluster during failure cases? if not we might consider putting this all in a single `oci_exec_test` with a `trap`, since you're not relying on the `helm_release` readiness signal anyways.

my only concern is leaving dangling clusters during flakey tests. in CI it doesn't really matter, but can be a real pain locally.

joshrwolf: additionally, that might make it easier in the future to use the cilium cli to "properly" check readiness with `cilium status --wait`

images/cilium/tests/main.tf
tcnghia: agent image is coming

images/cilium/tests/main.tf
tcnghia: I like the idea of using `cilium` CLI to use their `cilium connectivity test` as well. However, in case of failed test on CI it is very hard to know what's going on without Pods log uploaded (post-submit flake is even harder).

Locally, this pain is felt more by the people working on the image itself and someone who may rebuild everything, but it is probably a good trade off at least in the beginning when we haven't flushed out all the flakes. I'd still move everything to a single script, but I'll keep the cluster in case of failures.

tcnghia: Done

PR #1663, opened 2023-10-17
Pin pulumi to Python 3.11

tcnghia:
Current grpcio dep version doesn't work on 3.12.

### DIFF
--- a/images/pulumi/configs/latest.apko.yaml
+++ b/images/pulumi/configs/latest.apko.yaml
@@ -15,9 +15,9 @@ contents:
     - aspnet-7-targeting-pack
     # for Pulumi Python support
     - pulumi-language-python
-    - python3
-    - python3-dev
-    - py3-pip
+    - python-3.11
+    - python-3.11-dev
+    - py3.11-pip
     # for Pulumi Node.js support
     - pulumi-language-nodejs
     - nodejs
--- a/images/pulumi/configs/latest.apko.yaml
+++ b/images/pulumi/configs/latest.apko.yaml
@@ -15,7 +15,7 @@ contents:
     - aspnet-7-targeting-pack
     # for Pulumi Python support
     - pulumi-language-python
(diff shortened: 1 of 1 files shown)


images/pulumi/configs/latest.apko.yaml
```diff
@@ -15,9 +15,9 @@ contents:
     - aspnet-7-targeting-pack
     # for Pulumi Python support
     - pulumi-language-python
-    - python3
-    - python3-dev
-    - py3-pip
+    - python-3.11
```
k4leung4, line 18: add a comment to explain why its pinned to 3.11

tcnghia, line 18:
```suggestion
    # Currently pulumi has a dependency on grpcio version 1.56, which does not work with 3.11
    # See https://github.com/grpc/grpc/issues/33063.
    - python-3.11
```

images/pulumi/configs/latest.apko.yaml
```diff
@@ -15,9 +15,11 @@ contents:
     - aspnet-7-targeting-pack
     # for Pulumi Python support
     - pulumi-language-python
-    - python3
-    - python3-dev
-    - py3-pip
+    # Currently pulumi has a dependency on grpcio version 1.56, which does not work with 3.11
```
k4leung4, line 18: did you mean not work for 3.12?

tcnghia, line 18:
```suggestion
    # Currently pulumi has a dependency on grpcio version 1.56, which does not work with 3.12
```

images/pulumi/configs/latest.apko.yaml
tcnghia: Done

k4leung4 (approved):

tcnghia:
@k4leung4 I'm waiting to hear from @kaniini first to merge. Thanks for the review!

PR #1666, opened 2023-10-18
cilium-test: inherit the pull secret from the active context

tcnghia:
On CI, the active context K3s test cluster may contain additional pull secrets that we need to propagate to the cilium-test-specific cluster as well.


images/cilium/tests/cilium-install.sh
```diff
@@ -14,29 +15,41 @@ if [ -z "$CI" ]; then
     trap "k3d cluster delete $CLUSTER_NAME" EXIT
 fi
 
+TMPDIR=$(mktemp -d --tmpdir=$SCRIPT_DIR)
+trap "rm -rf $TMPDIR" EXIT
```
mattmoor, line 19: I thought there could only be on trap, so I'm a little concerned that this will leak k3s clusters.

tcnghia: I've consolidated into one trap. Thanks

joshrwolf (approved):

PR #1694, opened 2023-10-19
refactors spire image

mritunjaysharma394:
### Image Size

- [ ] The Image is smaller in size than its common public counterpart.
- [ ] The Image is larger in size than its common public counterpart (please explain in the notes).

Notes:

### Image Vulnerabilities

- [ ] The Grype vulnerability scan returned 0 CVE(s).
- [ ] The Grype vulnerability scan returned > 0 CVE(s) (please explain in the notes).

Notes:

- [ ] The container image was successfully loaded into a kind cluster.
- [ ] The container image could not be loaded into a kind cluster (please explain in the notes).

Notes:

- [ ] The application is accessible to the user/cluster/etc. after start-up.
- [ ] The application is not accessible to the user/cluster/etc. after start-up. (please explain in the notes).

Notes:

### Helm

- [ ] A Helm chart has been provided and the container image can be used with the chart.  If needed, please add a -compat package to close any gaps with the public helm chart.
- [ ] A Helm chart has been provided and the container image is not working with the chart (please explain in the notes).
- [ ] A Helm chart was not provided.

Notes:

### Processor Architectures

- [ ] The image was built and tested for x86_64.
- [ ] The image could not be built for x86_64 (please explain in the notes).
- [ ] The image was built and tested for aarch64.
- [ ] The image could not be built for aarch64. (please explain in the notes).

Notes:

- [ ] Functional tests have been included and the tests are passing.  All tests have been documnted in the notes section.

Notes:

- [ ] There has not been a request and/or there is no indication that this image needs tested on a public cloud provider.
- [ ] The container image has been tested successfully on a public cloud provider (AWS, GCP, Azure).
- [ ] The container image has not been tested successfully on a public cloud provider (AWS, GCP, Azure) (please explain in the notes).

Notes:

### Version

- [ ] The package version is the latest version of the package.  The latest tag points to this version.
- [ ] The package version is the not the latest version of the package (please explain in the notes).

Notes:

### Dev Tag Availability

- [ ] There is not a dev tag available that includes a shell and apk tools (by depending on 'wolfi-base') (please explain in the notes).

Notes:

- [ ] The image requires a non-standard username or non-standard GID/UID (please explain in the notes).

### ENTRYPOINT

- [ ] applications/servers/utilities set to call main program with no arguments e.g. [redis-server]
- [ ] applications/servers/utilities not set to call main program with no arguments e.g. [redis-server] (please explain in the notes)
- [ ] base images leave empty.
- [ ] base image and not empty (please explain in the notes).
- [ ] dev variants is set to entrypoint script that falls back to system.
- [ ] dev variants is not set to entrypoint script that falls back to system (please explain in the notes).

### CMD

### Environment Variables

- [ ] Environment variables added.
- [ ] Environment variables not added and not required.

### SIGTERM

- [ ] The image responds to SIGTERM (e.g., `docker kill $(docker run -d --rm cgr.dev/chainguard/nginx)`)

### Logs

- [ ] Error logs write to stderr and normal logs to stdout. Logs DO NOT write to file.

### Documentation - README

- [ ] A README file has been provided and it follows the README template.


images/spire/config/main.tf
```diff
@@ -11,9 +11,14 @@ variable "extra_packages" {
...
   type        = list(string)
   default     = []
+
 }
 
-module "accts" { source = "../../../tflib/accts" }
+module "accts" {
+  source = "../../../tflib/accts"
+  # agent runs as root while others as non-root, this condition is for that
+  run-as = var.name == "agent" ? 0 : 65532
```
stormqueen1990, line 20: Any reason why not make it a variable here?
For example,
```terraform
variable "run-as" {
  description = "The user with which this should run as"
}
```

Then this line would be:
```terraform
  run-as = var.run-as
```

images/spire/config/main.tf
```diff
@@ -22,7 +27,8 @@ output "config" {
     }
     accounts = module.accts.block
     entrypoint = {
-      command = "/usr/bin/spire-${var.name}"
+      ## agent and server require `run` while oidc-discovery-provider doesn't, this condition tries to handle that
+      command = var.name == "agent" || var.name == "server" ? "/usr/bin/spire-${var.name} run" : "/usr/bin/${var.name}"
```
stormqueen1990, line 31: Following the same idea from the previous comment:
```terraform
  command = var.command
```

images/spire/main.tf
```diff
@@ -1,40 +1,40 @@
...
 
-module "latest" {
-  for_each          = local.components
-  source            = "../../tflib/publisher"
-  name              = basename(path.module)
-  target_repository = "${var.target_repository}-${each.key}"
-  config            = file("${path.module}/configs/latest.${each.key}.apko.yaml")
-  build-dev         = true
+locals {
+  components = toset(["server", "agent", "oidc-discovery-provider"])
```
stormqueen1990, line 12: Can this be made into a map where you also carry your `run-as` and `command`, and then send it into the `config` via variables as explained in the previous comments?

images/spire/config/main.tf
mritunjaysharma394: Perfect, will do, thanks so much!

images/spire/main.tf
mritunjaysharma394: Not sure if I understood this part, do we have any similar examples for it?

stormqueen1990: I don't know if we currently have anything similar. What I had in mind was something around the lines of:
```terraform
locals {
  component_data = {
    "server" = {
      run-as  = 0
      command = "server"
    },
    // other components here
  }
  components = toset(keys(local.component_data))
}
```
so you won't need conditionals in the `config` module.

mritunjaysharma394: Thanks @stormqueen1990 , just tried to refactor with your amazing suggestions, feel free to take a look

mritunjaysharma394: it worked I guess 🥳

images/spire/config/main.tf
```diff
@@ -0,0 +1,43 @@
...
+
+# agent runs as root while others as non-root, this variable is for that
+variable "run-as" {
+  description = "The user with which this should run as"
+  default     = 65532
+}
+
+# agent and server require `run` while oidc-discovery-provider doesn't, this variable tries to handle that
+variable "command" {
+  description = "The user with which this should run as"
```
stormqueen1990, line 25: Seems like this was copypasted from the previous variable!

```suggestion
  description = "The command to be used as the entrypoint"
```

images/spire/config/main.tf
```diff
@@ -0,0 +1,43 @@
...
+  default     = ""
```
stormqueen1990, line 26: How about leaving this without a default? I think it'd be better to error out if this is something we expect to be present than to default as empty, which is more difficult to debug if forgotten.

```suggestion
```

images/spire/config/main.tf
```diff
@@ -0,0 +1,43 @@
...
+}
+module "accts" {
+  source = "../../../tflib/accts"
+  run-as = var.run-as
+}
+
+output "config" {
+  value = jsonencode({
+    contents = {
+      packages = concat(["spire-${var.name}"], var.extra_packages)
```
stormqueen1990, line 36: As per `images/spire/configs/latest.agent.apko.yaml`, the `agent` component also needed `libcap-utils` and `busybox` before, but I don't see those being added here.
Are these packages not needed anymore?

images/spire/configs/latest.oidc-discovery-provider.apko.yaml
```diff
@@ -1,18 +0,0 @@
...
-  users:
-    - username: nonroot
-      uid: 65532
-      gid: 65532
-  run-as: 65532
-
-entrypoint:
-  command: /usr/bin/oidc-discovery-provider
-
-cmd: --help
```
stormqueen1990, line 18: This is missing in the new structure.

images/spire/config/main.tf
mritunjaysharma394: Thanks @stormqueen1990 , just fixed it!

images/spire/config/main.tf
mritunjaysharma394: Just attempted to fix it, thanks!

images/spire/config/main.tf
```diff
@@ -0,0 +1,48 @@
...
+    }
+    accounts = module.accts.block
+    entrypoint = {
+      command = var.command
+    }
+    cmd = var.cmd
```
stormqueen1990, line 46: This could be done in a follow-up PR but I think we should avoid using `command` and `cmd` because these can become confusing. Perhaps `entrypoint-command` would be a better choice here.

images/spire/config/main.tf
```diff
@@ -0,0 +1,48 @@
...
+
+# agent and server require `run` while oidc-discovery-provider doesn't, this variable tries to handle that
+variable "command" {
+  description = "The command to be used as the entrypoint"
+}
+
+# oidc-discovery-provider begins wiht --help command after entrypoint, this is to handle that
+variable "cmd" {
+  description = "The command followed by entrypoint"
+}
```
stormqueen1990, line 31: This one could have a default of empty, since it's not mandatory for this specific module.

```suggestion
  default = ""
}

```

images/spire/main.tf
```diff
@@ -1,40 +1,67 @@
...
-  config            = file("${path.module}/configs/latest.${each.key}.apko.yaml")
-  build-dev         = true
+
+locals {
+  component_data = {
+    "server" = {
+      run-as         = 65532
+      command        = "/usr/bin/spire-server run"
+      extra_packages = []
+      cmd            = ""
```
stormqueen1990, line 18: If `cmd` also has a default, then this is unneeded.

```suggestion
```

images/spire/main.tf
```diff
@@ -1,40 +1,67 @@
...
+    },
+
+    "agent" = {
+      run-as         = 0
+      command        = "/usr/bin/spire-agent run"
+      extra_packages = ["busybox", "libcap-utils"]
+      cmd            = ""
```
stormqueen1990, line 25: Similar to the previous one.

```suggestion
```

images/spire/main.tf
```diff
@@ -1,40 +1,67 @@
...
+    },
+
+    "oidc-discovery-provider" = {
+      run-as         = 65532
+      command        = "/usr/bin/oidc-discovery-provider"
+      extra_packages = []
```
stormqueen1990, line 31: This defaults to empty, so no need to specify.
```suggestion
```

stormqueen1990 (approved):
Added some cleanup suggestions, but feel free to address them in a follow-up PR.

PR #1705, opened 2023-10-20
Refactor wait-for-it to config module

mattmoor:


images/wait-for-it/config/main.tf
```diff
@@ -0,0 +1,19 @@
...
+  default     = ["wait-for-it"]
+}
+
+data "apko_config" "this" {
+  config_contents = file("${path.module}/template.apko.yaml")
+  extra_packages  = var.extra_packages
+}
+
+output "config" {
+  value = jsonencode(data.apko_config.this.config)
```
k4leung4, line 18: is there a preference for inlining the config versus using a templated yaml?

k4leung4 (approved):

images/wait-for-it/config/main.tf
mattmoor: Generally I prefer the pure HCL because it enables better shared libraries and refactoring (e.g. `accts`) but this is definitely the simplest migration path.

PR #1714, opened 2023-10-20
Bug fix: Run fluentd-bit as root user

priyawadhwa:
To properly read log files, it looks like fluent-bit needs to run as the root user. The open source project has some issues open to allow running as nonroot, but for now it doesn't look like that's possible:
* https://github.com/fluent/helm-charts/issues/330
* https://github.com/aws/aws-for-fluent-bit/issues/427
* https://github.com/kube-logging/logging-operator/issues/972

Ideally we'd be able to run as nonroot, but I don't see a way around this for now 🤔 

should fix https://github.com/chainguard-dev/image-quality-issues/issues/47


images/fluent-bit/tests/main.tf
```diff
@@ -27,6 +27,8 @@ resource "helm_release" "fluent-bit" {
       repository = data.oci_string.ref.registry_repo
       tag        = data.oci_string.ref.pseudo_tag
     }
+
+    command = ["/usr/bin/fluent-bit"]
```
mattmoor, line 31: Is this needed?  It’s already the entrypoint

images/fluent-bit/config/template.apko.yaml
```diff
@@ -10,7 +10,7 @@ accounts:
   users:
     - username: nonroot
       uid: 65532
-  run-as: 65532
+  run-as: 0
```
vaikas, line 13: nit: maybe add your findings in the excellent PR description to an issue, then point to it from here?

mattmoor (dismissed):

vaikas (dismissed):

images/fluent-bit/config/template.apko.yaml
priyawadhwa: done!

images/fluent-bit/tests/main.tf
priyawadhwa: so the helm chart rewrites the entrypoint to `/fluent-bit/bin/fluent-bit` so we need to explicitly set it to the path in our image

images/fluent-bit/config/template.apko.yaml
mattmoor: Hmm, the issue is private and this is in a public repo 😞

mattmoor: Can we include a `tl;dr` inline maybe?

images/fluent-bit/tests/main.tf
mattmoor: Can you just throw that in a comment for now, so it's clear later?  I also started a meta-thread on how we document the knobs we need for Helm compat for users (vs. just our tests!)

mattmoor (approved):
Couple places we could use comments, but otherwise LGTM. 🙏

images/fluent-bit/tests/main.tf
priyawadhwa: yep, added instructions in the README for using the helm chart as well.

images/fluent-bit/config/template.apko.yaml
priyawadhwa: Ah good catch 😓 I ended up linking to this PR description

priyawadhwa:
@mattmoor thanks, it should be fixed now! I also realized we had an `aws-for-fluent-bit` image which also needs to be updated to run as the root user.

mattmoor (approved):

PR #1727, opened 2023-10-23
refactor gradle to config format

imjasonh:


images/gradle/config/main.tf
```diff
@@ -0,0 +1,28 @@
...
+
+data "apko_config" "this" {
+  config_contents = file("${path.module}/template.apko.yaml")
+  extra_packages  = var.extra_packages
+}
+
+output "config" {
+  value = jsonencode(merge(data.apko_config.this.config, {
+    environment : {
+      "JAVA_HOME" : var.java_home
```
k4leung4, line 25: JAVA_HOME is still defined in template.apko.yaml on line 24
keep one of them?

imjasonh: Good catch, removed from the YAML

k4leung4 (approved):

PR #1730, opened 2023-10-23
fix: nvidia-device-plugin + manual steps taken for testing, also adds…

rawlingsj:
… a helm variant

The manual testing steps we took here are described in the README.md below, the tests use a helm chart which may or may not be used by users however that chart requires bash and for the image to run as root.  Therefore this PR also creates a `latest-helm` tag variant so that users can pick the right image for how they use the image.

If there's other suggestions or better ways let me know.

Shout out to @k4leung4 for the manual testing steps - that was not easy.


rawlingsj:
Going to drop the helm-latest variant and switch the main image to run as root as we expect nvidia gpu will need that.

PR #1743, opened 2023-10-24
istio test: don't install CRD validation webhook

tcnghia:
This is a cluster resource with non-customizable name. It interferes with running multiple Istio tests in parallel in the same cluster.

See https://github.com/istio/istio/blob/master/manifests/charts/base/templates/default.yaml

### DIFF
--- a/images/istio/tests/main.tf
+++ b/images/istio/tests/main.tf
@@ -76,6 +76,8 @@ resource "helm_release" "base" {
     global = {
       istioNamespace = local.namespace
     }
+    # Disable the CRD validation webhook to avoid contention w/ tests of other versions,
+    # as this is a cluster-wide resource that we can't customize the name.
     defaultRevision = ""
   })]
 }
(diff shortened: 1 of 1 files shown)


images/istio/tests/main.tf
```diff
@@ -76,6 +76,7 @@ resource "helm_release" "base" {
     global = {
       istioNamespace = local.namespace
     }
+    defaultRevision = ""
```
k4leung4, line 79: can you add a comment to say you are disabling the install of CRD validation webhook, due to conflict with other tests

tcnghia, line 79:
```suggestion
    # Disable the CRD validation webhook to avoid contention w/ tests of other versions,
    # as this is a cluster-wide resource that we can't customize the name.
    defaultRevision = ""
```

tcnghia: Done

rawlingsj (approved):

PR #1752, opened 2023-10-24
fix argocd:latest(-dev) tagging

imjasonh:
Previously we'd been building updates to both argocd and argocd-repo-server into the same repository (argocd-repo-server).


found-it (approved):

mattmoor (approved):
Is there a repo we need to withdraw?

PR #1767, opened 2023-10-25
adds minor nit fixes in spire image

mritunjaysharma394:
just some minor nits as suggested by @stormqueen1990 in #1694

### DIFF
--- a/images/spire/config/main.tf
+++ b/images/spire/config/main.tf
@@ -21,13 +21,14 @@ variable "run-as" {
 }
 
 # agent and server require `run` while oidc-discovery-provider doesn't, this variable tries to handle that
-variable "command" {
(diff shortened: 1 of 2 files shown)


images/spire/config/main.tf
```diff
@@ -43,6 +44,6 @@ output "config" {
     entrypoint = {
       command = var.command
     }
-    cmd = var.cmd
+    cmd = var.entrypoint-cmd
```
stormqueen1990, line 47: This should be the other way around. It's `var.command` that goes to the entrypoint bit, so that variable should be called `entrypoint-*` instead of this one.

mritunjaysharma394: Oh okay thanks for this! I got confused, just fixing it

PR #1772, opened 2023-10-26
monopod scaffold: generate contents from template dir

imjasonh:
This refactors `monopod scaffold` to walk a tree of possibly-templated files in `images/TEMPLATE` and generate the same tree with inputs.

This also moves `-m` to a positional arg, e.g. `monopod scaffold blahh`

Example generated output from running `monopod scaffold blahh` here: https://github.com/chainguard-images/images/compare/main...imjasonh:images:scaffold-example?expand=1


monopod/pkg/commands/scaffold.go
```diff
@@ -69,15 +45,15 @@ func Scaffold() *cobra.Command {
...
   monopod scaffold --package-name test --entrypoint /usr/bin/test --run-as 65530 --user-gid 65534 --group-gid 65534`,
 
-		RunE: func(_ *cobra.Command, _ []string) error {
-			return o.runScaffold()
+		RunE: func(cmd *cobra.Command, args []string) error {
+			o.ModuleName = args[0]
+			return o.runScaffold(cmd.Context())
 		},
-		Args: cobra.NoArgs,
+		Args: cobra.ExactArgs(1),
```
vaikas, line 52: nit, but we should probably update the documentation above to specify that you must specify the module-name.

monopod/pkg/commands/scaffold.go
```diff
@@ -91,73 +67,70 @@ func Scaffold() *cobra.Command {
...
-
-	if err := o.scaffoldMainTerraform(mainTfOutfile); err != nil {
-		return err
-	}
-
-	// this file should be written in the configs directory
-	mainConfigTfOutfile, err := o.createOutputFile(filepath.Join(ConfigFolder, outputMappings[MainConfigTfTemplate]))
-	if err != nil {
-		return err
+	if err := os.MkdirAll("images/"+o.ModuleName, os.ModePerm); err != nil {
```
vaikas, line 75: I like the mkdirall, but in this case I'm not sure we should to prevent users from running it in the incorrect locations. I only say because I know of previous instances (and not only me) where some tools were not ran from the correct locations, just a thought. At least I think there's quite a few assumptions about the location of the code, and from both repos I think having the images directory there should always be the case, and not being there is a pretty good signal that there's a problem?

imjasonh: That's a good point. If you run this from another directory (e.g., inside `images/apko`), it will fail with

```
Error: failed to check templates folder at path images/TEMPLATE/: stat images/TEMPLATE/: no such file or directory
```

...which isn't great, but at least it's something, and can hopefully guide people to run it in the right place.

vaikas: Perfect! As long as it fails then it's all great. Thanks for the explanation.

vaikas (approved):
Woot Woot!

images/TEMPLATE/tests/main.tf
```diff
@@ -0,0 +1,40 @@
...
+  namespace        = "{{.ModuleName}}-${random_pet.suffix.id}"
+  repository       = "" // TODO
+  chart            = "" // TODO
+  create_namespace = true
+
+  values = [
+    jsonencode({
+      // TODO:
+    })
+  ]
```
stormqueen1990, line 33: Is there a reason why this would be preferred over `set{}`?

monopod/pkg/commands/scaffold.go
```diff
@@ -91,192 +67,70 @@ func Scaffold() *cobra.Command {
 
 // runScaffold does the real work of creating the folders, files, and evaluating
 // templates.
-func (o scaffoldOptions) runScaffold() error {
+func (o scaffoldOptions) runScaffold(ctx context.Context) error {
```
stormqueen1990, line 70: This context doesn't look like it's in use right now. Is it a good practice to always have it, even when not being leveraged?

images/TEMPLATE/tests/main.tf
imjasonh: I don't really know, I mostly just copied this from another helm_release and slapped a TODO on it.

If `set` is easier we can make that the default too, very easily.

monopod/pkg/commands/scaffold.go
imjasonh: You're right. I'd piped it through out of habit, and then forgot to use it for anything. 

I'd say it's nice to have it already piped-through, even if you're not using it, just so there's one less excuse to use `context.Background()` if you need a context later.

I don't consider it a hard fast rule though.

vaikas: Yeah, having it wired through is nice. I'd change the `ctx` to `_` though?

stormqueen1990: Agree with having it as `_`.

images/TEMPLATE/tests/main.tf
stormqueen1990: Mostly curious, but I don't think any change to this is warranted.

monopod/pkg/commands/scaffold.go
stormqueen1990, line 70:
```suggestion
func (o scaffoldOptions) runScaffold(_ context.Context) error {
```

imjasonh: I think I'd prefer to keep it called `ctx`, so it's clearer that someone doesn't need `context.Background()`

PR #1778, opened 2023-10-27
cilium: bump the test cluster nodes

tcnghia:
Use at least 2 nodes to exercise cross-node connectivity.


imjasonh (approved):

tcnghia:
the `curl` test timed out once but didn't happen again. I'll monitor and bump timeout if this becomes a problem in post-submit

PR #1793, opened 2023-10-30
feat: add new make targets and documentation

stormqueen1990:
Add new `make` targets to run Terraform. This uses the recommended `terraform apply` command that was previously documented in `TERRAFORM.md`. It also updates the `TERRAFORM.md` to suggest the use of `make` now and specifies how variables should be passed in to make.


jonjohnsonjr (approved):

stormqueen1990:
This depends on the fix in #1795 to be merged.

PR #1796, opened 2023-10-30
Add environment variables to nvidia-device-plugin image

found-it:
The nvidia-device-plugin image was not loading the correct libraries when using the upstream chart at https://nvidia.github.io/k8s-device-plugin, it seems it's missing these environment variables.


images/nvidia-device-plugin/config/main.tf
```diff
@@ -9,11 +9,25 @@ variable "extra_packages" {
...
+    }
+    accounts = module.accts.block
+    entrypoint = {
+      command = "/usr/bin/nvidia-device-plugin"
+    }
+    environment = {
+      "NVIDIA_VISIBLE_DEVICES" : "all",
+      "NVIDIA_DRIVER_CAPABILITIES" : "utility",
+      "NVIDIA_DISABLE_REQUIRE" : "true",
+      "NVIDIA_MIG_MONITOR_DEVICES" : "all",
```
k4leung4, line 30: just curious, where does this env come from?

i only find reference of it coming from the helm chart.
https://github.com/search?q=repo%3ANVIDIA%2Fk8s-device-plugin%20NVIDIA_MIG_MONITOR_DEVICES&type=code

images/nvidia-device-plugin/config/main.tf
```diff
@@ -9,11 +9,25 @@ variable "extra_packages" {
...
+  value = jsonencode({
+    contents = {
+      packages = var.extra_packages
+    }
+    accounts = module.accts.block
+    entrypoint = {
+      command = "/usr/bin/nvidia-device-plugin"
+    }
+    environment = {
+      "NVIDIA_VISIBLE_DEVICES" : "all",
```
k4leung4, line 27: can you leave a comment that we should mirror the env from https://github.com/NVIDIA/k8s-device-plugin/blob/main/deployments/container/Dockerfile.ubi8#L55C1-L57C39

images/nvidia-device-plugin/config/main.tf
found-it: Ah yep you're right, I was going off of a running container but this seems to be set by the helm chart. Will remove so it can be configured properly

images/nvidia-device-plugin/config/main.tf
found-it: Done! Good suggestion

images/nvidia-device-plugin/config/main.tf
found-it: Done

k4leung4 (approved):

PR #1797, opened 2023-10-31
WIP: use opentofu

imjasonh:


imjasonh:
Need to get our providers to be trusted by opentofu, can't make progress until then.

PR #1801, opened 2023-10-31
Reduce `APKO_IMAGE` usage, add default

mattmoor:


imjasonh (approved):

mattmoor:
To expand a little bit on the motivation for this change...  There are two versions that have to be kept in-sync for the reproducibility check to pass.

The first is the `tf-apko` provider version , which we version in the module here: https://github.com/chainguard-images/images/blob/dbb9e9f1cc83b8c8cea30cf950579f0886cb0b02/tflib/publisher/main.tf#L7-L10

The second is the digest, which this PR moves alongside that version within the publisher module in the `check-reproducilibity.sh` script.

These two were already co-versioned in our downstream repo, but here we made the `APKO_IMAGE` variable so that we could test some of the real-world image builds _**while overriding the `tf-apko` version**_, which necessitated being able to override the `APKO_IMAGE` to keep the versions consistent so the reproducibility check passed.

However, when I introduced `APKO_IMAGE` I did it such that it had to be overridden everywhere (not just in `tf-apko`) and this meant these two versions, which MUST be kept in sync were no longer co-versioned because `APKO_IMAGE` was defined in an actions workflow file in the respective repos, which meant that downstream it is impossible to atomically update the two versions together.

PR #1823, opened 2023-11-04
Adding additional test to validate keycloak startup in production mode

mamccorm:
We already have a helm test which runs keycloak in dev mode. This introduces a new shell test which runs Keycloak in production mode. This requires a keystore to be created and volume mounted.

### DIFF
--- a/images/keycloak/tests/main.tf
+++ b/images/keycloak/tests/main.tf
@@ -44,7 +44,7 @@ resource "helm_release" "test" {
 
 data "oci_exec_test" "keycloak-production-test" {
   digest = var.digest
-  script = "${path.module}/01-keycloak-production-mode.sh"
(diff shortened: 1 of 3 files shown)


images/keycloak/tests/main.tf
```diff
@@ -42,6 +42,11 @@ resource "helm_release" "test" {
   })]
 }
 
+data "oci_exec_test" "keycloak-production-test" {
+  digest = var.digest
+  script = "${path.module}/01-keycloak-production-mode.sh"
```
imjasonh, line 47: The numbering scheme was only necessary under a different testing strategy, you can just name this without the 01.

mamccorm: Good to know, thanks for calling out @imjasonh - amended as suggested :)

imjasonh (approved):

PR #1835, opened 2023-11-08
feat: improve Makefile checks

stormqueen1990:
Improve the way make targets check for mandatory variables and ensure the `apko-build` and `apko-build-alpine` targets do not require the Terraform variables to be set to run.

Also remove the re-assignment of Terraform variables, as those are unneeded.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,17 @@ ifeq ($(TF_AUTO_APPROVE),1)
 TF_VARS += --auto-approve
 endif
 
+.PHONY: check-env-tf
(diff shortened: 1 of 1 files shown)


imjasonh:
I think the apko-build targets are increasingly going to be less useful, as some configs get migrated to TF completely.

If it's easier to just remove them that may end up happening eventually.

Not requiring the TF var makes sense though.

imjasonh (dismissed):

jonjohnsonjr (approved):

PR #1842, opened 2023-11-08
disable aws-lb tests until they're good

imjasonh:
The helm-cleanup doesn't seem to cleanly remove the webhook, which means other installations running in the same cluster get blocked on webhook errors.

### DIFF
--- a/images/aws-load-balancer-controller/tests/main.tf
+++ b/images/aws-load-balancer-controller/tests/main.tf
@@ -11,10 +11,10 @@ variable "digest" {
 
 data "oci_string" "ref" { input = var.digest }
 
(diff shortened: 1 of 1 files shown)


images/aws-load-balancer-controller/tests/main.tf
```diff
@@ -13,6 +13,9 @@ data "oci_string" "ref" { input = var.digest }
 
 resource "random_pet" "suffix" {}
```
stormqueen1990, line 14: This could be a part of the commented block.

```suggestion
// resource "random_pet" "suffix" {}
```

stormqueen1990 (approved):

PR #1848, opened 2023-11-09
fix: pin Pulumi to the image version

stormqueen1990:
Pin the version of the Pulumi Python package to the same version that's currently being used in the image to avoid issues with divergent configurations.

This intends to fix the test that's currently failing on CI for us.

### DIFF
--- a/images/pulumi/examples/smoketest-python/requirements.txt
+++ b/images/pulumi/examples/smoketest-python/requirements.txt
@@ -1,2 +1,2 @@
-pulumi
+pulumi==3.91.1
 pulumi-kubernetes


tcnghia (approved):

imjasonh (approved):

imjasonh:
This is fine to unblock, but we should unpin so that when 3.91.2 comes out we're not back in the same situation.

stormqueen1990:
> This is fine to unblock, but we should unpin so that when 3.91.2 comes out we're not back in the same situation.

Yes. Our package in wolfi-dev/os should probably be bumped to the latest 3.92.0.

PR #1871, opened 2023-11-16
give cert-manager tests more time to happen and teardown

imjasonh:
Running multiple of these on a crowded cluster makes this take longer.


jonjohnsonjr (approved):
image

PR #1888, opened 2023-11-24
Add metadata.hcl and updated README.md files for all images

jamonation:
This PR does not include changes to monopod. It only adds a `metadata.hcl` file for each image to track things like the URL of a project, end of life data, a short description for use on Academy and images console etc.

There is also an `README.md` for each image based on the provided metadata, and rendered from a template in monopod. 

The monopod changes to render things will be in a separate PR since it needs everyone to switch to using it.


images/argo/metadata.hcl
```diff
@@ -0,0 +1,9 @@
+name                = "argo"
+image               = "cgr.dev/chainguard/argo"
+logo                = "https://storage.googleapis.com/chainguard-academy/logos/argo.svg"
+endoflife           = ""
+console_summary     = ""
+short_description   = "Images for working with [Argo workflows](https://argoproj.github.io/argo-workflows/)"
+compatibility_notes = ""
+readme_file         = "README.md"
```
imjasonh, line 8: Will this ever not be README.md?

imjasonh (approved):
Love this! We should make this part of `monopod scaffold` too, by creating a metadata.hcl.tmpl in TEMPLATE

images/argo/metadata.hcl
jamonation: I wasn't sure about this - thinking about the future where it might be a set of files depending on the different images that things like cert-manager might create.

But maybe it would be better to just make it a constant for now and we can revisit a future monopod if this needs to change?

jamonation:
> Love this! We should make this part of `monopod scaffold` too, by creating a metadata.hcl.tmpl in TEMPLATE

Glad you thought of that too, I added it over in #1889: [monopod/pkg/commands/templates/metadata.hcl.tpl](https://github.com/chainguard-images/images/blob/684f967ed3be8162aca03709b0ae60745d01b0f3/monopod/pkg/commands/templates/metadata.hcl.tpl)

jamonation:
So close here, but testing with `autodocs` seems to result in no readme body content at all, so marking this as a draft until I can figure out why the content is getting eaten.

PR #1903, opened 2023-11-30
Set minio image to use minio user

denhamparry:
- minio image requires minio user

### DIFF
--- a/images/minio/config/main.tf
+++ b/images/minio/config/main.tf
@@ -6,7 +6,13 @@ variable "entrypoint" {
   description = "The entrypoint to use for the image"
 }
 
-module "accts" { source = "../../../tflib/accts" }
+module "accts" {
+  source = "../../../tflib/accts"
+  name   = "minio"
+  uid    = 1000
+  gid    = 1000
+  run-as = 1000
+}
 
 output "config" {
   value = jsonencode({


images/minio/config/template.apko.yaml
```diff
@@ -0,0 +1,24 @@
+contents:
+  packages:
+    - minio
+    - minio-oci-entrypoint
+    - busybox
```
joshrwolf, line 5:
```suggestion
```

you can remove this since it's a package dependency now

denhamparry: I think I can remove both, as the entry script was just creating a user that we're already doing

images/minio/config/template.apko.yaml
```diff
@@ -0,0 +1,22 @@
...
+  groups:
+    - groupname: minio
+      gid: 1000
+  users:
+    - username: minio
+      uid: 1000
+  run-as: 1000
+
+entrypoint:
+  command: /usr/bin/docker-entrypoint.sh
```
joshrwolf, line 15:
```suggestion
  command: /usr/bin/minio
```

images/minio/config/template.apko.yaml
```diff
@@ -0,0 +1,22 @@
...
+    - minio
+
+accounts:
+  groups:
+    - groupname: minio
+      gid: 1000
+  users:
+    - username: minio
+      uid: 1000
+  run-as: 1000
```
joshrwolf, line 12: did you pick this? or did it come from somewhere? not disagreeing just asking for context

joshrwolf:
leaving for future ref, upstream uses [this](https://explore.ggcr.dev/layers/quay.io/minio/minio@sha256:7b40c61317c661d5d87428a287cd3f9c588ad4177b3ae50107be040359837485/usr/bin/docker-entrypoint.sh) entrypoint, which appears to be deprecated but still left in the image for whatever reason.

this approach simply defines the user as `1000` in the oci config, and makes the necessary adjustments to the `apko` file to ensure that user exists.

images/minio/config/template.apko.yaml
denhamparry: https://github.com/minio/minio/blob/master/dockerscripts/docker-entrypoint.sh
I looked at the file and could see it setup the user with the uid and gid.
I stuck with it as I searched and saw some other images do it.

imjasonh (approved):

denhamparry:
Confirmation that this isn't required. Minio is working as expected so doesn't require this PR. Closing the PR.

Issue #1898, opened 2023-11-28
Postgres 16 client support with wolfi-base


jasaltvik:
### Which image/versions are related to this issue/feature request?

The latest `postgres` image (ish)

### Issue/Feature description

Hi,

I used the [postgres](https://github.com/chainguard-images/images/blob/main/images/postgres/config/main.tf#L9) image as inspiration to create a really simple Postgres 15 client image like this

```Dockerfile
FROM cgr.dev/chainguard/wolfi-base

RUN apk update && apk add --no-cache --update-cache glibc-locale-en busybox su-exec bash postgresql-15-client libpq-15

ENV DURATION=30m
CMD sleep ${DURATION}
```

It's just used as a way for devs to spin up a container that can be used to connect to a Postgres server, and has been working fine.

Then Postgres 16 was released, and I thought I would try to upgrade the client image to 16 as well

```Dockerfile
RUN apk update && apk add --no-cache --update-cache glibc-locale-en busybox su-exec bash postgresql-16-client libpq-16
```

This works, but when connecting to the server I get

```bash
~ $ psql -U <user>
psql: error: connection to server at "<server>" (<ip>), port 5432 failed: FATAL:  no PostgreSQL user name specified in startup packet
connection to server at "<server>" (<ip>), port 5432 failed: FATAL:  no PostgreSQL user name specified in startup packet
double free or corruption (out)
Aborted (core dumped)
```

I've not been able to understand the cause of this. Do you have any input on what could be missing to make it work with Postgres 16? 

If this issue doesn't belong here, feel free to close it.

Thanks in advance.

camden-xage:
@jasaltvik I'm experiencing a similar issue, but I'm trying to create a minimal postgres image to run as a server to reproduce without all of our extra "fluff" I'm struggling to get it to throw this error for some reason (e.g. basic postgres image with just a single username/password set up). Are you able to share how you are running your postgres server? Is it just a simple docker setup or something similar?

mattmoor:
Digging into this we found an issue with the way `provider_priority` was being handled (it gets dropped when generating the index) which was leading to an older version of `libcrypt1` being installed by the solver (`r7` vs. `r5` for the rest of the stuff from `glibc`).

We're fixing that regardless, but if you have an easy repro, I am curious whether installing `libcrypt1=2.38-r7` explicitly resolves this for you.

ref: https://github.com/chainguard-dev/melange/pull/878

camden-xage:
[...]

PR #1933, opened 2023-12-11
add postgres-helm-compat and modify postgres tests

joshrwolf:
adds a new `postgres-helm-compat` image intended to be compatible with the popular helm based postgresql installations (which is currently the bitnami helm chart).

This also removes our existing `postgres` helm chart based test, which is incompatible with the `postgres` image.


found-it:
Should we add a `docker` based `tls` test for the regular postgres image?

joshrwolf:
> Should we add a `docker` based `tls` test for the regular postgres image?

yes, but at the risk of giving a cop out. I think that test should belong at the package level instead of here

lmk if you disagree, if not I'll do a follow up in wolf/ios

found-it:
I was thinking about this a bit today @joshrwolf . I'm still on the fence about whether the TLS test should land in the package itself. My one concern is about _when_ we run package tests. If we only run them at package build time, I don't think we would actually catch an issue like [this one](https://github.com/chainguard-images/images/issues/1898#issuecomment-1842092883). In that issue, the postgres package didn't change - it was the openssl version in the image that changed and broke the application.

Please correct me if I'm wrong but I don't think we have a way to re-run package tests based on changing dependencies. If we had a mechanism to re-run package tests when certain dependencies change then I think that will solve our issue since it includes both prebuilt images and adding the package on a wolfi-base. Otherwise right now I believe testing the image is really the only way we could catch an issue with dependencies drifting over time.

To sum up, at the risk of asking you to do even more work - my vote is still in the image. I think there's value in having a package test to make use of as we evolve package testing but I don't know if it would catch what we need it to right now.

found-it:
Cool, yeah I like that for now so we can block updating the postgres image if there are TLS issues. Thanks for adding the test! Love the use of the k8s job btw

found-it (approved):

PR #1943, opened 2023-12-13
Upgrade publisher to run NTIA conformance checker.

mattmoor:
I expect this to fail some places, but let's see how bad it is...


mattmoor:
Let's see how much closer we get if all of the dev variants are fixed 🤞 
https://github.com/wolfi-dev/os/pull/9892

imjasonh (approved):

PR #1944, opened 2023-12-13
add tctl-authorization-plugin subpackage

shivanshu1333:
add package


vaikas (approved):
Once the package gets built, this should work.

PR #1970, opened 2023-12-19
temporarily disable check-sbom where it's enabled

imjasonh:
This blocks some downstream checks where the SBOMs we're producing aren't fully valid yet.


joshrwolf (approved):

mattmoor:
Do we have a tracking issue for the underlying issue with context?

PR #1971, opened 2023-12-19
Improvements to Go + Node Image READMEs

SharpRake:
This PR contains updates to the READMEs for the `node` and `go` images. Fingers crossed that the new workflow doesn't cause issues with these updates.

Looking for DevEn approval before this should be merged, thank you!

### DIFF
--- a/images/node/README.md
+++ b/images/node/README.md
@@ -35,41 +35,41 @@ It specifies `NODE_PORT=3000` by default.
 <!--compatibility:end-->
 
 <!--body:start-->
-## Usage Example
+## Node Application Example
 
-Navigate to the [`example/`](https://github.com/chainguard-images/images/tree/main/images/node/example) directory:
+This brief example is derived from our [Getting Started with Node](https://edu.chainguard.dev/chainguard/chainguard-images/getting-started/getting-started-node/) guide which is itself based on the [Docker Node example](https://docs.docker.com/language/nodejs/containerize/).
 
+Here, we'll assume you have a node app that you want to containerize and run. You can check out [Step 1 of the Getting Started with Node guide](https://edu.chainguard.dev/chainguard/chainguard-images/getting-started/getting-started-node/#step-1-setting-up-a-demo-application) for instructions on setting up an example Node app.
+
+Navigate to your application directory. Ensure your Dockerfile uses the Chainguard `node` base image.
+
+```sh
+nano Dockerfile
 ```
-cd example/
-```
-
-The Dockerfile is based on Docker's [Build your Node image](https://docs.docker.com/language/nodejs/build-images/) tutorial, but uses the Chainguard node base image instead.
-
-Build the application on the node base image.
-
-```
-docker build \
-  --tag node-docker \
-  --platform=linux/amd64 \
-  .
+```Dockerfile
+FROM cgr.dev/chainguard/node:latest
+... 
 ```
 
-Then you can run the image:
+Next, build the application. Note that this example tags the resulting image with `node-docker`.
 
-```
-docker run \
-  --rm -it \
-  -p 8000:8000 \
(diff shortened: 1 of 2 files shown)


images/node/README.md
```diff
@@ -35,41 +35,41 @@ It specifies `NODE_PORT=3000` by default.
...
-## Usage Example
+## Node Application Example
 
-Navigate to the [`example/`](https://github.com/chainguard-images/images/tree/main/images/node/example) directory:
+This brief example is derived from our [Getting Started with Node](https://edu.chainguard.dev/chainguard/chainguard-images/getting-started/getting-started-node/) guide which is itself based on the [Docker Node example](https://docs.docker.com/language/nodejs/containerize/).
 
-```
-cd example/
-```
+Here, we'll assume you have a node app that you want to containerize and run. You can check out [Step 1 of the Getting Started with Node guide](https://edu.chainguard.dev/chainguard/chainguard-images/getting-started/getting-started-node/#step-1-setting-up-a-demo-application) for instructions on setting up an example Node app.
```
ltagliaferri, line 42:
```suggestion
Here, we'll assume you have a Node app that you want to containerize and run. You can check out [Step 1 of the Getting Started with Node guide](https://edu.chainguard.dev/chainguard/chainguard-images/getting-started/getting-started-node/#step-1-setting-up-a-demo-application) for instructions on setting up an example Node app.
```

ltagliaferri (commented):
I wonder if it's worthwhile to bring in the Node Getting Started example into the Node image directory, like we have for the Go image? I think it would make for a better experience to try out the image. 

I mentioned in a Slack comment that the first line comment in the Go Dockerfile caused an error for me on macos for some reason 🤔

SharpRake:
> I wonder if it's worthwhile to bring in the Node Getting Started example into the Node image directory, like we have for the Go image? I think it would make for a better experience to try out the image.
> 
> I mentioned in a Slack comment that the first line comment in the Go Dockerfile caused an error for me on macos for some reason 🤔

I've resolved the Go issue by removing the troublesome comment. and I updated the EDU demo repo to include the node example. I also reframed the Node README to focus more on this example (rather than attempting to make it generally applicable to any Node application) to provide a more uniform experience.

images/node/README.md
```diff
@@ -35,41 +35,104 @@ It specifies `NODE_PORT=3000` by default.
 <!--compatibility:end-->
 
 <!--body:start-->
-## Usage Example
+## Node Application Example
```
mamccorm, line 38: [Non-blocking comments]
Appreciate you working to improve our docs :) From a consumer there is quite a lot of info here to digest, and it's not the easiest to follow.

Just a suggestion, but would it be cleaner to show them an example of how to use our image to run an existing node application? i.e heres an example of how to use our image with an existing node application.

We could link to the accademy and mention it for more verbose docs on how to build and run node applications?

ltagliaferri: Thanks @mamccorm, I think we have not started doing this yet, but we were planning to do some segmentation of the readmes through Jamon's Terraform work. @SharpRake can touch base with @jamonation about the recommendations there, but maybe this would prove a solution to have a front and center TL;DR that can live on the READMEs and also the Directory, and then have a longer-form example on Academy. Here is the doc that outlines this: https://docs.google.com/document/d/1QpaYcHLafizSyjoNSahZfybeESPEaWDNbcI4oWqM4nY/edit

SharpRake: +1 to what Lisa said!

I think for now we should publish these changes to get them out the door but I won't mark [my issue for the node README](https://github.com/chainguard-dev/edu/issues/1214) as done until I've connected with Jamon and got a plan together about how we can approach this better.

mamccorm (approved):

PR #1972, opened 2023-12-20
DNM: don't reuse the previous locked config to build dev variants

imjasonh:
We seem to be producing invalid solutions with our current approach.

Before this change, if an image requests packages [A, B, C] and we lock that to [A@1, B@2, C@3], then add a dev variant that requests [A@1, B@2, C@3, D, E], the result can be invalid according to `apk add A B C D E`. The solution only works because apko's solver is different (read: incorrect), and we're working to make it less different (read: correct) in the future.

In general we shouldn't tighten constraints like this, and should instead retain a copy of the original loose input request and augment _that_ with further constraints (like dev variants). This approximates that, by passing the original request (which is [itself already resolved](https://github.com/chainguard-images/images/blob/b4dcab56cfddbe3a5358a6645e4cd213df59c94f/images/crane/config/main.tf#L12-L19)!) to both the original and dev variants.

To resolve this completely we should pass the unresolved (but post-`extra_packages` mutation) config to the module, where it can get resolved and dev-variant'ed inside the tflib/publisher module.

Opening as draft to discuss and iterate, and see if anything breaks as a result of just this.


jdolitsky:

PR #1996, opened 2023-12-28
Update vector tests to allow them to be invoked from another module

mamccorm:
Tweaks to the vector terraform tests - to allow them to be invoked from other modules and be re-used. This includes support for specifying an alternative git tag for the helm tests.

### DIFF
--- a/images/vector/tests/main.tf
+++ b/images/vector/tests/main.tf
@@ -8,25 +8,36 @@ variable "digest" {
   description = "The image digest to run tests over."
 }
 
+variable "helm_chart_tag" {
+  description = "Whether to pull a specific version of the helm chart for testing."
+  default     = ""
+}
+
 data "oci_string" "ref" { input = var.digest }
 
+resource "random_id" "hex" { byte_length = 4 }
+
 resource "helm_release" "vector" {
-  name       = "vector"
-  repository = "https://helm.vector.dev"
-  chart      = "vector"
+  name              = "vector-${random_id.hex.hex}"
+  repository        = "https://helm.vector.dev"
+  chart             = "vector"
+  create_namespace  = true
+  namespace         = "vector-${random_id.hex.hex}"
+  timeout           = 120
 
   values = [jsonencode({
     localpv = {
       image = {
         registry   = join("", [data.oci_string.ref.registry, "/"])
         repository = data.oci_string.ref.repo
-        tag        = data.oci_string.ref.pseudo_tag
+        tag        = var.helm_chart_tag != "" ? var.helm_chart_tag : data.oci_string.ref.pseudo_tag
       }
     }
   })]
 }
 
 module "helm_cleanup" {
-  source = "../../../tflib/helm-cleanup"
-  name   = helm_release.vector.id
+  source    = "../../../tflib/helm-cleanup"
+  name      = helm_release.vector.id
+  namespace = helm_release.vector.namespace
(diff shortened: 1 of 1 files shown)


images/vector/tests/main.tf
```diff
@@ -8,25 +8,36 @@ variable "digest" {
...
+  namespace        = "vector-${random_id.hex.hex}"
+  timeout          = 120
 
   values = [jsonencode({
     localpv = {
       image = {
         registry   = join("", [data.oci_string.ref.registry, "/"])
         repository = data.oci_string.ref.repo
-        tag        = data.oci_string.ref.pseudo_tag
+        tag        = var.helm_chart_tag != "" ? var.helm_chart_tag : data.oci_string.ref.pseudo_tag
```
mattmoor, line 33: Hmm, I don't believe this will work.

mattmoor: Consider: we don't publish a tag until tests (including this) have completed, so the only times this would actually qualify the image in question would be if the tag incidentally applied to the same image that we'd previously let through.

mamccorm: Thanks @mattmoor, is your feedback in relation to this specifically (i.e before my amendments):

```bash
tag = data.oci_string.ref.pseudo_tag
```

Also bump @srishtih

mamccorm: ACK thanks Matt for calling out. I think in other such examples we use the digest, so applied the same to here

developer-guy (approved):

images/vector/tests/main.tf
```diff
@@ -10,23 +10,29 @@ variable "digest" {
...
+  namespace        = "vector-${random_id.hex.hex}"
+  timeout          = 120
 
   values = [jsonencode({
     localpv = {
       image = {
         registry   = join("", [data.oci_string.ref.registry, "/"])
         repository = data.oci_string.ref.repo
-        tag        = data.oci_string.ref.pseudo_tag
+        tag        = trimprefix(data.oci_string.ref.digest, "sha256:")
```
mattmoor, line 28: I'm sort of surprised that this works either TBQH.  Maybe it's some quirk of the helm chart that it does, but I'm guessing that this is deploying an `ImagePullBackoff` `vector` image to the k3s cluster and we just don't realize it because we don't try to use the `vector` instance while it's up.

There really isn't a `tag` that we can use here because there aren't any tags applied to the digest until tests have passed.  `pseudo_tag` is a game we play where we know that when the syntax `:foo@bar` is used, the `foo` part isn't used by anything, but it lets us tunnel a digest through tag fields in charts.

Perhaps the problem here is that `localpv` isn't actually something that appears in the `vector` Helm chart, so none of these values are actually being used 🤔 
https://github.com/vectordotdev/helm-charts/blob/00c9a2cc17cfe41a378db86d046227d24df04559/charts/vector/values.yaml#L26-L40

mattmoor: I think that the `trimprefix` you have is right, if you change `tag =` to `sha =`, but my confidence that the `localpv` piece is right is close to zero (if I'm looking at the right helm chart).

srishtih: I just checked this helm installation details. localpv shouldn't have been there, unsure why I added it previously. The tests passed because they used the upstream image.

mattmoor (approved):

PR #1999, opened 2023-12-28
git: refactor to config modules

imjasonh:
Building on #1997 this refactors the git images to the config module pattern.

This manages to share the config between alpine and wolfi-based images, since they're so similar. Alpine adds an extra APK repository to use, to find `git-lfs`.

Root/nonroot behavior is maintained as well, and tagging is unaffected.


images/git/alpine.tf
```diff
@@ -6,11 +13,9 @@ module "latest-alpine" {
   source            = "../../tflib/publisher"
   name              = basename(path.module)
   target_repository = var.target_repository
-  config            = file("${path.module}/configs/latest.alpine.${each.key}.apko.yaml")
+  config            = module.alpine[each.key].config
   check-sbom        = false # Alpine-based SBOMs are not conformant because the Alpine baselayout has an invalid license specifier
-  # Override the module's default wolfi packages that conflict with alpine
-  extra_packages = []
```
mattmoor, line 12: Do we not still need this?

mattmoor (approved):
Other than the concern around `extra_packages` this LGTM!

images/git/alpine.tf
imjasonh: The default is `[]` already, it shouldn't have any effect (please fact check)

mattmoor: You are right, this must predate our use of the provider for this.

PR #1958, opened 2023-12-15
shard presubmit builds

imjasonh:

PR #2005, opened 2023-12-29
shard release workflow

imjasonh:
Similar to #1958 

Today we run the scheduled release with each image sharded onto its own hosted runner. We're approaching the shard limit of 256 (currently 241), so instead we'll randomly shard the builds into 4 shards (~60 images per shard) that run on the 64-core GH runner instance.

Randomly sharding helps prevent noisy neighbors that tend to be near each other alphabetically (`crossplane` and `crossplane-aws` and `crossplane-azure`).

Smaller shards would lower the impact of noisy neighbors, and could let us run on smaller hosted runners. Larger shards lower build times, since subsequent builds can share the same APK cache, and k3s cluster spin-up time can be amortized.

This also removes the last use of `generate-matrix` -- after this lands, we can drop that action and `monopod matrix` which is called by it.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -24,7 +24,7 @@ env:
 
 jobs:
   shard:
-    runs-on: ubuntu-latest-64-cores
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
 
@@ -65,7 +65,7 @@ jobs:
       matrix: "${{steps.shard.outputs.matrix}}"
 
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-64-cores
     needs: shard
     strategy:
       fail-fast: false
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -90,8 +90,8 @@ jobs:
         with:
           # This allows chainguard-images/images-private to publish images to cgr.dev/chainguard-private
           # We maintain this identity here:
-          # https://github.com/chainguard-dev/mono/blob/main/env/chainguard-images/iac/images-private-pusher.tf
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -66,76 +80,58 @@ jobs:
...
-
       - uses: chainguard-dev/actions/setup-chainctl@main
         with:
-          # This allows chainguard-images/images to publish images to cgr.dev/chainguard
+          # This allows chainguard-images/images-private to publish images to cgr.dev/chainguard-private
           # We maintain this identity here:
-          # https://github.com/chainguard-dev/mono/blob/main/env/chainguard-images/iac/images-pusher.tf
-          identity: 720909c9f5279097d847ad02a2f24ba8f59de36a/b6461e99e132298f
+          # https://github.com/chainguard-dev/mono/blob/main/env/chainguard-images/iac/images-private-pusher.tf
+          identity: ce2d1984a010471142503340d670612d63ffb9f6/ac92e3a8b3865440
```
mattmoor, line 94: This is wrong

.github/workflows/release.yaml
```diff
@@ -15,48 +15,62 @@ on:
...
+          remainder=$((total % n))
+
+          declare -a bins
+          # Sequentially fill up each bin, and append any remainders to the last bin
+          for ((i = 0; i < total; i++)); do
+            idx=$((i < (total - remainder) ? i / base_size : n - 1))
+            bins[$idx]+="${images[$i]} "
+          done
+
+          matrix=$(printf "%s\n" "${bins[@]}" | jq -cRnjr '[inputs] | [ range(0; length) as $i | { "index": $i | tostring, "images": .[$i] } ]')
```
jdolitsky, line 51: oh... my..... 🤯

jdolitsky (approved):

PR #2035, opened 2024-01-03
Add script to cleanup CRDs

mamccorm:
Helm was leaving behind CRDs created as part of a test. Adding cleanup script to ensure no data is left behind when the test concludes

### DIFF
--- a/images/prometheus-config-reloader/tests/main.tf
+++ b/images/prometheus-config-reloader/tests/main.tf
@@ -38,9 +38,24 @@ resource "helm_release" "kube-prometheus-stack" {
   }
 }
 
-module "helm_cleanup" {
+data "oci_exec_test" "helm_cleanup" {
+  digest = var.digest
+  script = "${path.module}/cleanup.sh"
+
+  env {
+    name  = "CHART_NAME"
+    value = helm_release.kube-prometheus-stack.name
+  }
+  env {
+    name  = "NAMESPACE"
+    value = helm_release.kube-prometheus-stack.namespace
+  }
+}
+
+
+/* module "helm_cleanup" {
   source    = "../../../tflib/helm-cleanup"
   name      = helm_release.kube-prometheus-stack.id
   namespace = helm_release.kube-prometheus-stack.namespace
-}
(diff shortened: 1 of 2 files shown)


images/prometheus-config-reloader/tests/main.tf
```diff
@@ -38,9 +38,24 @@ resource "helm_release" "kube-prometheus-stack" {
   }
 }
 
-module "helm_cleanup" {
+data "oci_exec_test" "helm_cleanup" {
+  digest = var.digest
```
developer-guy, line 42: there should be _depends_on_ section to wait until the helm command succeed I guess to provide the dependency between resources.

mamccorm: Thats a valid suggestion @developer-guy, thanks for that, will include and push up

images/prometheus-config-reloader/tests/main.tf
```diff
@@ -38,9 +38,17 @@ resource "helm_release" "kube-prometheus-stack" {
   }
 }
 
-module "helm_cleanup" {
-  source    = "../../../tflib/helm-cleanup"
-  name      = helm_release.kube-prometheus-stack.id
-  namespace = helm_release.kube-prometheus-stack.namespace
+data "oci_exec_test" "helm_cleanup" {
```
imjasonh, line 41: Worth noting this still requires the test to be serialized, which means using the `flock` trick we use with tigera-operator's tests.

It might just make sense to only test one variant, and skip it for others, or use the `flock` trick and serialize them and clean them up fully after each run.

@joshrwolf has ideas in the works about how to make this all _much_ better and more hermetic, but until then we only have growing piles of band-aids :-/

mamccorm: ACK, thought we just needed the cleanup part but makes sense, was hoping to not have to move the helm test into a shell script, but if thats the only option then I guess we can. Might ask @mritunjaysharma394 to close out this part tomorrow

mattmoor (approved):

PR #2050, opened 2024-01-05
updating JDK + JRE READMEs

SharpRake:
Quick update to improve the content of the JDK and JRE READMEs. Both READMEs follow the same example, so other than one sentenace calling out the specific Image these two READMEs should be identical.


sheesh:
LGTM from a dev enablement perspective.

imjasonh (approved):

PR #2059, opened 2024-01-10
isolate cert-manager tests using imagetest provider

joshrwolf:
Introduce and use [`imagetest`](https://github.com/chainguard-dev/terraform-provider-imagetest) for `cert-manager`.

This creates an ephemeral `k3s` cluster and uses the accompanying sandboxed test harness to a) install cert-manager via the helm chart and b) tickle it with `cmctl`.

Unlike the prior tests, these don't use the shared "host" cluster, and thus don't have the same conflict issues. Meaning they can be run in parallel for multiple instantiations of the root `module.cert-manager`.

Follow up PR will include this for the remaining tests that depend on `flock`, but I wanted to keep this PR small and focused


joshrwolf:
this introduces a `tflib/imagetest` module, which triggers our build the world.

to attempt to make this easier to review, here is the `cert-manager` related success: https://github.com/chainguard-images/images/actions/runs/7701846442/job/20988886736?pr=2059#step:7:2950

there's a whole class of a ~dozen images that fail due to various race conditions with `cert-manager` existing or not existing on the current shared cluster. rather than whack-a-mole with those, I'm hoping we can land this and iterate once there's no longer a "global" `cert-manager`

jonjohnsonjr (approved):

main.tf
```diff
@@ -12,6 +13,25 @@ terraform {
...
+      networks = {
+        // wire in k3d's default network where the registry lives
+        "k3d-default" = { name = "k3d-k3s-default" }
+      }
+      registries = {
+        # Mirror the var.target_repository host registry to the local registry.
+        # This ensures the images that are pushed from the host registry are
+        # mirrored to the internal hostname:port registry.
+        "${element(split("/", var.target_repository), 0)}" = {
+          mirror = { endpoints = ["http://${element(split(":", var.target_repository), 0)}:5000"] }
```
stormqueen1990, line 28: Does this work when the sandbox network has not been connected to the existing registry network?

found-it:
Love it

main.tf
joshrwolf: good catch! 

turns out when you spin up the registry with the same cluster create command, it attaches the registry to the cluster network (`k3s-k3d-default`).

since CI also spins up the registry with the same `host:port` combo, the only thing we need here is the network attachment

stormqueen1990 (approved):

main.tf
```diff
@@ -12,6 +13,25 @@ terraform {
   backend "inmem" {}
 }
 
+provider "imagetest" {
+  harnesses = {
```
joshrwolf, line 17: as written, this approach adds the necessary provider config to the `main.tf` to make things work for CI presubmit, as well as for folks using `k3d` locally.

however, this will **fail** for local dev when `k3d-k3s-default` is not used, which isn't the best UX.

an alternative would be to create this in a `main_override.tf` within the presubmit job so that this only applies to presubmit CI.

tflib/imagetest/helm/main.tf
```diff
@@ -0,0 +1,60 @@
+variable "name" {
```
imjasonh, line 1: nit, not blocking: might be worth splitting into variables.tf if we have this many

tflib/imagetest/helm/main.tf
```diff
@@ -0,0 +1,60 @@
...
+  --wait \
+  --timeout ${var.timeout}
+  EOb
+}
+
+output "install_cmd" {
+  value = local.install_cmd
+}
+
+output "uninstall_cmd" {
```
imjasonh, line 58: Where are these cmds used?

imjasonh (approved):

imjasonh (approved):

imjasonh:
Force merging because touching tflib rebuilds every image, and some of those images' tests are busted.

PR #2073, opened 2024-01-17
updating git, curl, and ruby image READMEs

SharpRake:
Another README update, this one updates the ruby, curl, and git Image READMEs. 

One thing I want to call out is that I left some unique formatting in the `git` README to call out the glibc version of the Image. The intent is that it looks like it's part of the "Get it!" section but if this is too wonky let me know and I'll adjust it.


images/curl/README.md
```diff
@@ -28,28 +28,26 @@ docker pull cgr.dev/chainguard/curl:latest
 <!--body:start-->
 ## Usage
 
-The curl image allows you to run ordinary curl commands in CI/CD pipelines and also locally via Docker.
+The curl Image allows you to run ordinary `curl` commands in CI/CD pipelines as well as locally with Docker.
```
ltagliaferri, line 31: I noticed the last batch had "x Chainguard Image"
```suggestion
The curl Chainguard Image allows you to run ordinary `curl` commands in CI/CD pipelines as well as locally with Docker.
```

ltagliaferri (commented):
LGTM, thank you!

erikaheidi (approved):

PR #2080, opened 2024-01-18
Revert "Revert "update Chainguard repo metadata with readme contents" "

imjasonh:
Before this change:

```
$ terraform plan -target=module.ko -var=target_repository=cgr.dev/chainguard
│ Error: Invalid index
│ 
│   on tflib/publisher/chainguard.tf line 25, in resource "chainguard_image_repo" "repo":
│   25:   parent_id = lookup(local.groups, local.group, data.chainguard_group.group[0].id)
│     ├────────────────
│     │ data.chainguard_group.group is empty tuple
│ 
│ The given key does not identify an element in this collection value: the collection has no elements.
```

It seems that `lookup(map, key, default)` tries to at least resolve `default` even if it doesn't have to use it, which in the case of `cgr.dev/chainguard/*` fails because the data source isn't invoked in that case.

This change gets around this by using `lookup(map, key, "") == "" ? default : map[key]`, which doesn't try to resolve `default` unless it's actually going to be used.


jdolitsky (commented):
do we need to include the module in the top-level main.tf ?

imjasonh:
> do we need to include the module in the top-level main.tf ?

I don't think so. That seems to only be needed when we configure the provider in that module, and we're just using the default provider config now.

PR #2082, opened 2024-01-18
Add dev variants to aspnet-runtime

jonjohnsonjr:


images/aspnet-runtime/main.tf
```diff
@@ -17,6 +17,7 @@ module "latest" {
 
   target_repository = var.target_repository
   config            = module.config.config
+  build-dev         = true
```
mattmoor, line 20: this needs an `oci_tag` for `latest-dev` below

imjasonh (approved):

PR #2087, opened 2024-01-19
scaffold: generate a metadata.hcl

imjasonh:
`monopod scaffold foo` doesn't create a metadata.hcl today, and I think it should.

Folks should still run `monopod readme` after editing that file, but at least this gets them started.


jdolitsky (commented):
doesnt this need some business in monopod to process it?

jamonation (approved):
Huh, dunno why I didn't think of this, thanks!

PR #2097, opened 2024-01-22
Make known Chainguard groups a variable

jdolitsky:


imjasonh (commented):
This might be painful to pass in through the root main.tf, since it has to go through each image module to get to tflib :-/

PR #2120, opened 2024-01-25
Best practices: Add section showing how to configure grype

pdeslaur:
I had to find the content of this configuration file recently. I decided to add it here, but we can move it around!


luhring (approved):
This looks great! We may want to add the config section that forces Grype to pull from the registry if left unspecified, just to avoid folks accidentally scanning stale images — but this is great regardless of if we want that

PR #2156, opened 2024-02-02
Add Apache spark (dependency), and adding tests

mamccorm:
The Spark operator has a dependency on spark. This PR adds the apache spark package to spark-operator. Also extended some testing.


images/spark-operator/tests/main.tf
```diff
@@ -9,26 +9,3 @@ variable "digest" {
 }
 
 data "oci_string" "ref" { input = var.digest }
-
-resource "helm_release" "spark-operator" {
```
mamccorm, line 13: Was removed to aid debugging. Will be added back.

PR #2160, opened 2024-02-03
Add Apache spark (dependency), and adding tests

mamccorm:
The Spark operator has a dependency on spark. This PR adds the apache spark package to spark-operator. Also extended some testing.

### DIFF
--- a/images/spark-operator/config/main.tf
+++ b/images/spark-operator/config/main.tf
@@ -1,17 +1,32 @@
 variable "extra_packages" {
   description = "Additional packages to install."
   type        = list(string)
-  default     = ["spark-operator"]
+  default     = ["spark", "spark-operator"]
 }
 
-module "accts" { source = "../../../tflib/accts" }
+module "accts" {
+  source = "../../../tflib/accts"
+}
 
+# TODO: procps should be moved to a runtime dependency of the spark-operator
+# package. At the time of writing, encountering issues loading runtime deps into
+# images.
 output "config" {
   value = jsonencode({
     contents = {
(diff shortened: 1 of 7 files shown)


images/spark-operator/config/main.tf
```diff
@@ -1,17 +1,32 @@
...
-  default     = ["spark-operator"]
+  default     = ["spark", "spark-operator"]
 }
 
-module "accts" { source = "../../../tflib/accts" }
+module "accts" {
+  source = "../../../tflib/accts"
+}
 
+# TODO: procps should be moved to a runtime dependency of the spark-operator
```
pdeslaur, line 11: Is this todo still applicable?

images/spark-operator/config/main.tf
```diff
@@ -1,17 +1,32 @@
 variable "extra_packages" {
   description = "Additional packages to install."
   type        = list(string)
-  default     = ["spark-operator"]
+  default     = ["spark", "spark-operator"]
```
pdeslaur, line 4: Can we move `spark` to the list of packages below? Otherwise it'll get overridden with just `spark-operator` inside the private build.

images/spark-operator/config/main.tf
mamccorm: Thanks for spotting - this wasn't required in the end - removed the package but not the comment

images/spark-operator/config/main.tf
mamccorm: Hey @pdeslaur, there are some circumstances where we need to override both packages, so this is intentional. It does mean both need to be passed in when / if this module is invoked elsewhere

pdeslaur (approved):

PR #2172, opened 2024-02-06
create a build job summary

joshrwolf:
uses a [gha job summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) to aggregate the test results

this is a primitive table to start, but we can flush this out further with links and details once the initial GHA madness is confirmed to work.

I tested this on a private repo, and kept this small, so I'm hopeful 🤞. either way, this doesn't touch the actual release process so the side effects are negligible, ci is always red nowadays anyways 😮‍💨

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -158,3 +158,32 @@ jobs:
             https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
 
             ${{ steps.tf-diag.outputs.errors }}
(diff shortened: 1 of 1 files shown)


joshrwolf:
here's an example of what this looks like. don't mind the duplicates, it's lazy test data 😅 

CleanShot 2024-02-06 at 09 22 28@2x

imjasonh (approved):
🥹

PR #2177, opened 2024-02-06
Reference local values and chart with helm module

found-it:
The imagetest helm module doesn't have a way to reference local charts and values files. Just adding that capability here

### DIFF
--- a/tflib/imagetest/helm/main.tf
+++ b/tflib/imagetest/helm/main.tf
@@ -1,23 +1,31 @@
 variable "name" {
-  default     = ""
   description = "The name of the helm release. One will be generated if not provided"
+  type        = string
+  default     = ""
+
 }
 
 variable "namespace" {
   description = "The namespace to install the helm release into."
+  type        = string
+  default     = "default"
 }
 
 variable "chart" {
   description = "The name of the helm chart to install."
+  type        = string
 }
 
 variable "repo" {
   description = "The helm repo to install the chart from."
+  type        = string
+  default     = ""
 }
 
 variable "chart_version" {
-  default     = ""
   description = "The version of the helm chart to install. The latest will be used if none is provided."
+  type        = string
+  default     = ""
 }
 
 variable "values" {
@@ -26,8 +34,15 @@ variable "values" {
   default     = {}
 }
 
+variable "values_file" {
+  description = "Path to values file on the local filesystem."
+  type        = string
+  default     = ""
+}
+
 variable "timeout" {
   description = "The timeout on the helm install."
+  type        = string
   default     = "5m"
 }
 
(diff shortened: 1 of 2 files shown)


tflib/imagetest/helm/main.tf
```diff
@@ -1,23 +1,30 @@
 variable "name" {
-  default     = ""
   description = "The name of the helm release. One will be generated if not provided"
+  type        = string
+  default     = ""
 }
 
 variable "namespace" {
   description = "The namespace to install the helm release into."
+  type        = string
```
joshrwolf, line 9: you've already done the work, but in the future you can 🪓 these, they're the default terraform type

tflib/imagetest/helm/main.tf
```diff
@@ -26,22 +33,30 @@ variable "values" {
...
 
 locals {
   install_cmd = <<-EOa
 apk add helm
 helm install ${var.name} ${var.chart} \
-  --namespace ${var.namespace} --create-namespace \
-  --repo ${var.repo} \
-  --wait --wait-for-jobs \
-  --timeout ${var.timeout} \
+  %{if var.namespace != "default"}--namespace ${var.namespace} --create-namespace%{endif} \
```
joshrwolf, line 52: what does this buy us over just passing in `var.namespace`? which is always going to be at _least_ `default`?

tflib/imagetest/helm/main.tf
```diff
@@ -1,23 +1,30 @@
...
 }
 
 variable "namespace" {
   description = "The namespace to install the helm release into."
+  type        = string
+  default     = "default"
 }
 
 variable "chart" {
   description = "The name of the helm chart to install."
```
joshrwolf, line 14:
```suggestion
  description = "The name of, or path to the helm chart to install."
```

tflib/imagetest/helm/main.tf
```diff
@@ -1,23 +1,30 @@
...
+  type        = string
 }
 
 variable "repo" {
   description = "The helm repo to install the chart from."
```
joshrwolf, line 19:
```suggestion
  description = "The helm repo to install the chart from. When not specified, a local helm chart will be assumed."
```

tflib/imagetest/helm/main.tf
```diff
@@ -26,22 +33,30 @@ variable "values" {
   default     = {}
 }
 
+variable "values_file" {
+  description = "Path to values file on the local filesystem."
+  type        = string
+  default     = ""
```
joshrwolf, line 39:
```suggestion
  type        = list(string)
  default     = []
```

imo we make this a list, since multiple `--values` is valid, and I can see us leveraging this in the future

tflib/imagetest/helm/main.tf
found-it: Meh I was going for succinctness for the inputs but in reality it doesn't actually change anything. Good call, it doesn't buy us anything

tflib/imagetest/helm/main.tf
found-it: Happy to axe them, terraformls was yelling at me that there was no type so I just added them in

tflib/imagetest/helm/main.tf
found-it: Agreed here, needs a couple other things to make it work though, will push a commit

found-it: Since we're going to allow multiple values files I'm going to make a note about the precedence. Highest precedence will be the inline values, then the last values file in the list, the lowest precedence will be the first values file specified

tflib/imagetest/helm/main.tf
found-it: axed

joshrwolf (approved):

joshrwolf:
I'm going to bypass branch protections on this one. the failures in CI are due to changes being made in `tflib/**`, but these changes are scoped to `tflib/imagetest`.

the CI failures are occurring in unrelated images not yet using `imagetest`.

PR #2178, opened 2024-02-06
convert spire to imagetest, add functional tests

joshrwolf:
convert `spire` to `imagetest` to deconflict our image tests, and also lets us use `helm test`, which for this rare case actually provides a solid smoke test!


images/spire/tests/main.tf
```diff
@@ -19,51 +19,87 @@ data "oci_string" "ref" {
...
+        tag        = data.oci_string.ref["oidc-discovery-provider"].pseudo_tag
+      }
+      config = {
+        acme = {
+          tosAccepted = true
+        }
+      }
+    }
+  }
+}
```
stormqueen1990, line 76: Might be a good use case for `jsondecode`.

stormqueen1990 (approved):

images/spire/tests/main.tf
joshrwolf: as discussed, going to revisit this in a follow up if/when necessary 🙏

PR #2183, opened 2024-02-07
Tflib imagetest/helm values file fix

found-it:
An empty `values_file` parameter would generate a bad helm command. This change just ignores the `--values` line corresponding to values files if the `values_file` parameter is omitted.

Converted the prometheus-logstash-exporter image tests over to use the helm module as a test


joshrwolf (approved):

joshrwolf:
force merging with similar logic to #2177

PR #2187, opened 2024-02-07
deconflict kyverno{-policy-reporter} using imagetest

joshrwolf:
deconflicts kyverno and kyverno-policy-reporter by isolating them on separate test harnesses. follows the similar model of #2174

### DIFF
--- a/images/kyverno/tests/helm/main.tf
+++ b/images/kyverno/tests/helm/main.tf
@@ -1,0 +1,65 @@
+variable "values" {
+  type = any
+  default = {
+    admissionController = {
+      container = {
+        image = {
+          registry   = "cgr.dev"
+          repository = "chainguard/kyverno"
+          tag        = "latest"
+        }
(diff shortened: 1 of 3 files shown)


joshrwolf:
I'm pleasantly surprised with how good the `helm test`s are for `kyverno`, but I'm not happy with the `kyverno-policy-reporters` lack of tests, which will need to be addressed in follow up PRs

images/kyverno/tests/helm/main.tf
stormqueen1990, line 1: I don't want to hold the PR up on that but using `jsondecode` here might be a good idea.

stormqueen1990 (approved):

images/kyverno/tests/helm/main.tf
joshrwolf: as discussed, going to revisit this in a follow up if/when necessary 🙏

PR #2190, opened 2024-02-08
migrate cluster-autoscaler to imagetests that actually do something

joshrwolf:
our existing `cluster-autoscaler` tests have never actually worked, we just happened to be sneaking them in because the startup readiness checks flap, this is to the point where its consistently broken ([ref](https://github.com/chainguard-images/images/actions/runs/7822870362))

in order to actually deploy the cluster-autoscaler chart in a way that works, it needs some provider to lock on to.

this changes the test to use `imagetest`, and uses `capi` as that "mock" provider for `cluster-autoscaler` to manage. we don't actually trigger a scale event, but just having `capi` on the cluster is enough for `cluster-autoscaler` to go through its regular readiness _and_ liveness sequence.

given that this is a static go app, I'm confident _enough_ that this should be sufficient without needing to fully stress the scale events against a real cloud provider

### DIFF
--- a/images/cluster-autoscaler/tests/main.tf
+++ b/images/cluster-autoscaler/tests/main.tf
@@ -46,7 +46,7 @@ resource "imagetest_feature" "basic" {
 
   steps = [
     {
-      name = "Convert to cluster to a CAPI management cluster"
+      name = "Convert cluster to a CAPI management cluster"
       cmd  = <<EOF
 apk add clusterctl
 clusterctl init --infrastructure docker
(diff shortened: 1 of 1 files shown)


images/cluster-autoscaler/tests/main.tf
```diff
@@ -40,11 +36,29 @@ resource "helm_release" "cluster-autoscaler" {
...
-  name      = helm_release.cluster-autoscaler.id
-  namespace = helm_release.cluster-autoscaler.namespace
+resource "imagetest_feature" "basic" {
+  harness     = imagetest_harness_k3s.this
+  name        = "Basic"
+  description = "Basic functionality of the cluster-autoscaler helm chart."
+
+  steps = [
+    {
+      name = "Convert to cluster to a CAPI management cluster"
```
found-it, line 49:
```suggestion
      name = "Convert cluster to a CAPI management cluster"
```

stormqueen1990 (approved):

PR #2200, opened 2024-02-09
WIP: Make opensearch deployable with upstream chart

found-it:
Taking a crack at getting opensearch functional with the upstream chart

### DIFF
--- a/images/opensearch/config/template.apko.yaml
+++ b/images/opensearch/config/template.apko.yaml
@@ -7,41 +7,41 @@ contents:
 accounts:
   groups:
     - groupname: opensearch
-      gid: 65532
+      gid: 1000
   users:
     - username: opensearch
-      uid: 65532
-  run-as: 65532
+      uid: 1000
+  run-as: 1000
 
 paths:
   - path: /usr/share/opensearch
     type: directory
-    uid: 65532
-    gid: 65532
+    uid: 1000
+    gid: 1000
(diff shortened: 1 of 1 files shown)


found-it:
Looks like upstream runs as user 1000 https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/values.yaml#L313

There's an initContainer with a hard-coded `chown 1000 ...` so we probably need to match upstream https://github.com/opensearch-project/helm-charts/blob/main/charts/opensearch/templates/statefulset.yaml#L249

found-it:
Unfortunately I did not get much farther this evening. The plugins are on disk in the upstream image but they aren't bundled in the tarball created by `./gradlew localDistro`. I tried `./gradlew assemble` as well and it seems they weren't bundled there either. [There's an `/opt/opensearch.tar.gz` ](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/docker/src/docker/Dockerfile#L48)in the Dockerfile that I think contains the plugins, but haven't been able to figure out where it's coming from to confirm. I think this will work once we have the plugins enabled, if we want to try just adding the plugins to simply test it out it seems there's an `installPlugins` list in the helm chart to install plugins in an init container. We should figure out how these things are getting on disk though. Happy to pick it up tomorrow

found-it:
Go ahead an merge this whenever!

PR #2203, opened 2024-02-09
emit more useful logs from failed imagetest helm installs

joshrwolf:
standard helm emits some pretty useless logs when it errors out:

https://github.com/chainguard-images/images-private/actions/runs/7845026704/job/21411213191#step:15:89

```
INSTALLATION FAILED: client rate limiter Wait returned an error: context deadline exceeded
```

This wraps the install in a block and captures the error, and prints a whole bunch more diagnostics similar to the `k8s-diag` error we use.

these will get captured by the provider and emitted as `diagnostics.details`, which in turn will show up in the CI logs as the `<detail></detail>`

### DIFF
--- a/tflib/imagetest/helm/main.tf
+++ b/tflib/imagetest/helm/main.tf
@@ -1,69 +1,45 @@
-variable "name" {
-  description = "The name of the helm release. One will be generated if not provided"
-  default     = ""
-}
-
-variable "namespace" {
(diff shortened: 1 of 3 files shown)


stormqueen1990 (approved):

joshrwolf:
I'm going to force merge this given the analysis above ^

PR #2216, opened 2024-02-12
chore: add option for `--create-namespace` in imagetest

stormqueen1990:
Include an option for specifying `--create-namespace` in imagetest. Default is kept as true.

### DIFF
--- a/tflib/imagetest/helm/main.tf
+++ b/tflib/imagetest/helm/main.tf
@@ -14,7 +14,8 @@ locals {
   install_cmd = <<-EOinstall
 apk add helm
 if ! helm install ${local.name} ${var.chart} \
-  --namespace ${var.namespace} --create-namespace \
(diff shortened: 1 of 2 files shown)


found-it (commented):
Just because I'm curious if you ran into a case for this, what does this buy us over keeping `--create-namespace` switched by default?

stormqueen1990:
> Just because I'm curious if you ran into a case for this, what does this buy us over keeping `--create-namespace` switched by default?

Right now, it is switched on by default and I don't want to break the use cases that rely on that. I need it to be false for the `opentelemetry-collector-contrib` tests, because there is a resource created in the namespace prior to the deployment of the Helm chart: https://github.com/chainguard-images/images/blob/f2ed8967ca8cd586fb3d375df6010cb0d4cb26a0/images/opentelemetry-collector-contrib/tests/main.tf#L53-L68

PR #2217, opened 2024-02-12
fix: opentelemetry-collector-contrib tests after 0.94.0 release

stormqueen1990:
### Changes
* Tests are failing since the release of v0.94.0. Add configuration that works with the new release.
* Migrate tests to the imagetest provider.

### DIFF
--- a/images/opentelemetry-collector-contrib/tests/main.tf
+++ b/images/opentelemetry-collector-contrib/tests/main.tf
@@ -1,8 +1,8 @@
 terraform {
   required_providers {
     oci        = { source = "chainguard-dev/oci" }
-    helm       = { source = "hashicorp/helm" }
     kubernetes = { source = "hashicorp/kubernetes" }
+    imagetest  = { source = "chainguard-dev/imagetest" }
   }
 }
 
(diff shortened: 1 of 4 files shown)


images/opentelemetry-collector-contrib/tests/main.tf
```diff
@@ -105,15 +120,52 @@ resource "helm_release" "open-telemetry-custom-config" {
...
+      name = "Create namespace and configmap with custom configuration for the deployment"
+      cmd  = <<EOF
+        NAMESPACE=otelc-deploy
+        kubectl create namespace $NAMESPACE
+        kubectl create configmap $NAMESPACE --namespace $NAMESPACE --from-file=/tests/custom-deploy-config.yaml
+      EOF
+    },
+    {
+      name = "Helm deploy opentelemetry-collector-contrib as a deployment"
+      cmd  = module.helm-otelc-deploy.install_cmd
```
found-it, line 152: Does this do verification on install or is there a way to verify it's working?

images/opentelemetry-collector-contrib/tests/main.tf
```diff
@@ -105,15 +120,52 @@ resource "helm_release" "open-telemetry-custom-config" {
...
+      name = "Create namespace and configmap with custom configuration for the daemonset"
+      cmd  = <<EOF
+        NAMESPACE=otelc-daemonset
+        kubectl create namespace $NAMESPACE
+        kubectl create configmap $NAMESPACE --namespace $NAMESPACE --from-file=/tests/custom-ds-config.yaml
+      EOF
+    },
+    {
+      name = "Helm deploy opentelemetry-collector-contrib as a daemonset"
+      cmd  = module.helm-otelc-daemonset.install_cmd
```
found-it, line 164: Does this do verification on install or is there a way to verify it's working?

stormqueen1990: Semantically this is the same test from before, just using imagetest instead.

found-it (approved):

PR #2219, opened 2024-02-13
enforce 5m timeout on temporal-server test

imjasonh:
I'm not sure if this is long enough, but it'd be nice to fail fast rather than time out after an hour, like we do now.


stormqueen1990:
Superseded by #2220

PR #2222, opened 2024-02-14
migrate cassandra suite of tests and rmqmto to imagetest

joshrwolf:
this migrates the cassandra suite of image tests (which all depend on `cert-manager`) to `imagetest` to prevent the conflicting cert-manager problem on HEAD.

as is, this is a **drop in** conversion to `imagetest`, meaning the bare minimum is done. this simplify involved moving the testing script over to the harness and running it in the feature.

in the future, we should probably improve these tests, as I've noticed a few of them don't really do much, when there's a whole [suite](https://docs.k8ssandra.io/quickstarts/site-reliability-engineer/) of things we can/should probably be testing.

cc @mritunjaysharma394 @developer-guy @srishtih since the vast majority of this work was done by you all!

additionally, I've also snuck in rmq mto to this PR, and hoisted the `cert-manager` installation into a reusable module


images/cass-config-builder/tests/smoke-test.sh
```diff
@@ -36,15 +34,14 @@ kubectl wait --for=condition=ready pod --selector app.kubernetes.io/name=k8ssand
 #reapply with new config
 current_config_builder=$(kubectl get configmap k8ssandra-operator-cass-operator-manager-config -n k8ssandra-operator -o jsonpath='{.data.image_config\.yaml}' | grep 'config-builder' | awk '{print $2}')
 echo $current_config_builder
-kubectl get configmap k8ssandra-operator-cass-operator-manager-config -n k8ssandra-operator -o yaml | \
-  sed "s|config-builder: ${current_config_builder}|config-builder: ${IMAGE_NAME}|" | \
-  kubectl apply -f -
+kubectl get configmap k8ssandra-operator-cass-operator-manager-config -n k8ssandra-operator -o yaml |
+	sed "s|config-builder: ${current_config_builder}|config-builder: ${IMAGE_NAME}|" |
```
stormqueen1990, line 38: I think you'll still need the `\` at the end of the line here, no?

joshrwolf: this was migrated verbatim from the existing tests

joshrwolf: it seems like they're effectively the same, the `\` after a `|` is purely for style reasons. TIL!

stormqueen1990: TIL too

stormqueen1990 (approved):

PR #2237, opened 2024-02-15
Fix test for kubeflow

pdeslaur:
The delay seems a bit too short. When running: `kubectl wait --for=condition=Ready --timeout=300s -n kubeflow --all pods`, there's a risk the command will fail right away with `error: no matching resources found`. I think increasing the delay of the retry helps here


luhring (approved):

joshrwolf:
do you have some logs for where this was failing with 

```
error: no matching resources found
```

the `retry` block in this new `imagetest` land allows for the command to fail a few times. this is the recommended way forward to handle these cases. adding the additional `1s` isn't really doing anything here (bad or good)

PR #2270, opened 2024-02-23
Print a list of images pushed to local registry on PR

jdolitsky:


imjasonh (commented):
Minor thing, but can we make this just a script somewhere, instead of making a make target? Make is weird.

jdolitsky:
@imjasonh done

PR #2278, opened 2024-02-27
move cassandra,envoy,etcd  to corresponding version streams

ajayk:
This needs https://github.com/wolfi-dev/os/pull/13738 to be merged first

### DIFF
--- a/images/cassandra/config/main.tf
+++ b/images/cassandra/config/main.tf
@@ -6,7 +6,7 @@ terraform {
 
 variable "extra_packages" {
   description = "The additional packages to install."
-  default     = ["cassandra", "cassandra-compat", "openjdk-11-default-jvm"]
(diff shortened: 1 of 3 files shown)


images/cassandra/config/main.tf
```diff
@@ -6,7 +6,7 @@ terraform {
 
 variable "extra_packages" {
   description = "The additional packages to install."
-  default     = ["cassandra", "cassandra-compat", "openjdk-11-default-jvm"]
+  default     = ["cassandra-4.1", "cassandra-4.1-compat", "openjdk-11-default-jvm"]
```
imjasonh, line 9: Will this need to be updated when 4.2+ comes out? We can potentially automate this, but it'd be better to have this not require edits when the packages change.

ajayk: This is not needed as the packages provides latest virtuals

PR #2284, opened 2024-02-28
add k8ssandra-system-logger

Dentrax:


images/k8ssandra-system-logger/README.md
```diff
@@ -0,0 +1,28 @@
...
+<!--getting:start-->
+## Download this Image
+The image is available on `cgr.dev`:
+
+```
+docker pull cgr.dev/chainguard/k8ssandra-system-logger:latest
+```
+<!--getting:end-->
+
+<!--body:start--><!--body:end-->
```
mamccorm, line 28: Can we add some more docs here on how to use the image? i.e perhaps with cass-operator helm chat which seems to be the most common? The chart does support multiple images (a lot of we also have), but the example could just show how to override just this single image with the chart.

If we think a docker example would be good too then we can also include

images/k8ssandra-system-logger/tests/01-test.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
mamccorm, line 1: Could we also add a helm test? (and use the same chart in our README docs). The good news is we should largely be able to re-use this:

- https://github.com/chainguard-images/images/blob/1c3b1504b7a54fbc62f9ba397fd3666fc4a62f50/images/cass-operator/tests/helm.sh

But with different image override to the cass-operator. I think this might be what we'd need to override:
- https://github.com/k8ssandra/k8ssandra/blob/726158d32fcdd4a33f9fd6ee8dc4d84042bc906b/charts/cass-operator/values.yaml#L82
-

images/k8ssandra-system-logger/config/main.tf
```diff
@@ -0,0 +1,19 @@
...
+    apko = { source = "chainguard-dev/apko" }
+  }
+}
+
+module "accts" { source = "../../../tflib/accts" }
+
+output "config" {
+  value = jsonencode({
+    contents = {
+      packages = ["cass-operator-config", "vector", "libstdc++"]
```
mamccorm, line 12: Given this particular image, i'm not sure if we'd ever need to switch out cass-operator-config, but perhaps to avoid needing to change in future, we can change this to the extra_packages approach?

i.e:

```
variable "extra_packages" {
  description = "The additional packages to install"
  default     = ["cass-operator-config"]
}
```

And then similarly in the config module:
```
packages  = concat(["libstdc++", "vector"], var.extra_packages)
```

mamccorm: Or is vector the main package here?

images/k8ssandra-system-logger/tests/01-test.sh
Dentrax: > Can we add some more docs here on how to use the image?

Sorry - I should have more careful on that. (re: previous review)

> i.e perhaps with cass-operator helm operator (see comment below in test section) which seems to be the most common?

> Could we also add a helm operator test?

> But with different image override to the cass-operator.

This image is _actually_ nothing more than a `vector` package and a configuration file. We do not use `cass-operator` here, but only the config version, which is: `cass-operator-config`. So I think that our smoke test would be totally OK in this case. FWIW, we should add some test to our `vector` package, as a follow-up item. WDYT?

images/k8ssandra-system-logger/README.md
Dentrax: Thanks, done!

images/k8ssandra-system-logger/config/main.tf
Dentrax: Correct, `vector` is the main package here.

> i'm not sure if we'd ever need to switch out cass-operator-config

The variables is compatible with the FIPS version of this image.

images/k8ssandra-system-logger/tests/01-test.sh
mamccorm: It looks like the cass-operator itself though would be used with this image, i.e how customers would consume this. So was more to test the image with the operator the way customers would use it.

From looking through the links above it looks like the cass-operator deploys system-logger, so we'd want to have a test for that added. Do let me know if i'm missing anything though :)

Dentrax:
This is currently blocked due to an issue on the CI. Upstream repo uses old version of CRDs:

```
Error: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "cass-operator-sunny-glider-grafana" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
ensure CRDs are installed first, resource mapping not found for name: "cass-operator-sunny-glider-grafana-test" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
```

PR currently on hold.

jdolitsky:

PR #2295, opened 2024-02-29
node: improve Dockerfile test, include dumb-init

imjasonh:
This moves our Dockerfile test to use a multi-stage build as described in [docker-node's best practices](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#smaller-images-without-npmyarn).

This proves the image doesn't _need_ a shell (or `npm`) to run basic apps in a best-practice way.

In the future we can use this test to ensure the image works as expected when/if we remove `sh` and `npm` from the image.

cc @bretfisher

### DIFF
--- a/images/node/example/Dockerfile
+++ b/images/node/example/Dockerfile
@@ -1,16 +1,19 @@
 # syntax=docker/dockerfile:1
 # Based on https://docs.docker.com/language/nodejs/build-images/
 
-ARG BASE=cgr.dev/chainguard/node
+ARG BUILDER=cgr.dev/chainguard/node:latest-dev
+ARG BASE=cgr.dev/chainguard/node:latest
+
+FROM ${BUILDER} AS builder
+WORKDIR /build-stage
+COPY package*.json ./
+ENV NODE_ENV=production
+RUN npm ci --omit=dev
+# Copy the the files you need
+COPY . ./
+# npm run build or whatever you need to do to build your app
 
 FROM ${BASE}
-
-ENV NODE_ENV=production
-
-COPY --chown=node:node ["package.json", "package-lock.json*", "./"]
-
-RUN npm install --production
-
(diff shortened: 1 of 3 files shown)


images/node/example/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "example",
   "version": "1.0.0",
   "description": "",
-  "main": "index.js",
+  "main": "server.js",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1"
```
smoser, line 7: This looks like it probably passes that string to `sh -c`.  which is really weird. but [doc](https://docs.npmjs.com/cli/v10/using-npm/scripts) mentions "command" and "execute".

smoser (commented):
This is nice. thank you.
I'm approving, but I do have one comment inline.

images/node/example/package.json
imjasonh: We could just get rid of this. This enables `npm run test`, which clearly isn't meant to succeed anyway.

smoser (approved):
thank you.

EyeCantCU (approved):

images/node/example/Dockerfile
```diff
@@ -1,16 +1,19 @@
 # syntax=docker/dockerfile:1
 # Based on https://docs.docker.com/language/nodejs/build-images/
 
-ARG BASE=cgr.dev/chainguard/node
-
-FROM ${BASE}
+ARG BUILDER=cgr.dev/chainguard/node:latest-dev
```
mattmoor, line 4: Should we be passing this ARG in any test modules?

mattmoor (approved):

images/node/example/Dockerfile
imjasonh: Not currently, but this will become useful for testing future changes.

PR #2309, opened 2024-03-04
docs(docker-selenium): Cover usage in README

EyeCantCU:
Adds a usage section to the README

### DIFF
--- a/images/docker-selenium/README.md
+++ b/images/docker-selenium/README.md
@@ -25,4 +25,78 @@ docker pull cgr.dev/chainguard/docker-selenium:latest
 ```
 <!--getting:end-->
 
-<!--body:start--><!--body:end-->
+<!--body:start-->
+
+## Usage
+
+Run the container with the following command:
+
+```bash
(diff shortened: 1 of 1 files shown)


images/docker-selenium/README.md
```diff
@@ -25,4 +25,73 @@ docker pull cgr.dev/chainguard/docker-selenium:latest
...
+driver.get("https://www.chainguard.dev/")
+sleep(5)
+
+'''Exit'''
+driver.close()
+driver.quit()
+print("ChromeDriver test successful!")
+```
+
+Please note that Chromium is ran in a headless state with the sandbox and GPU access disabled with the flags:
```
mamccorm, line 89: Can we look at expanding this section a little with the reasoning? i.e the reason we need to run in sandbox and with GPU disabled, being a container.. etc

EyeCantCU: Definitely!

mamccorm (approved):

PR #2312, opened 2024-03-04
Migrate k8ssandra-operator tests to imagetest

found-it:
k8ssandra-operator uses cert-manager which is causing test flakes

### DIFF
--- a/images/k8ssandra-operator/tests/main.tf
+++ b/images/k8ssandra-operator/tests/main.tf
@@ -1,7 +1,6 @@
 terraform {
   required_providers {
     oci       = { source = "chainguard-dev/oci" }
-    helm      = { source = "hashicorp/helm" }
     imagetest = { source = "chainguard-dev/imagetest" }
   }
 }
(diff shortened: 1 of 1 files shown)


images/k8ssandra-operator/tests/main.tf
```diff
@@ -1,63 +1,79 @@
 terraform {
   required_providers {
-    oci = { source = "chainguard-dev/oci" }
+    oci       = { source = "chainguard-dev/oci" }
+    helm      = { source = "hashicorp/helm" }
```
stormqueen1990, line 4: We can omit this

```suggestion
```

images/k8ssandra-operator/tests/main.tf
```diff
@@ -1,63 +1,79 @@
...
-    name  = "installCRDs"
-    value = "true"
+  sandbox = {
+    mounts = [
+      {
+        source      = path.module
+        destination = "/tests"
+      }
+    ]
   }
```
stormqueen1990, line 30: This block doesn't appear to be in use:
```suggestion
```

stormqueen1990 (approved):

PR #2322, opened 2024-03-07
Update dotnet README

erikaheidi:
Update .NET image README with correct pulling instructions. This PR addresses issue reported by @ryryryano

### DIFF
--- a/images/dotnet/README.md
+++ b/images/dotnet/README.md
@@ -18,6 +18,15 @@ Minimal images for .NET and the .NET Tools.
 
 <!--getting:start-->
 ## Download this Image
+The image is available on `cgr.dev`:
+
+```
+docker pull cgr.dev/chainguard/dotnet:latest
+```
+<!--getting:end-->
+
+<!--compatibility:start-->
+## Compatibility Notes
 
 The Chainguard .NET images are available on `cgr.dev` as two variants: `dotnet-sdk` and `dotnet-runtime`. The SDK variant contains additional tooling to facilitate development and building, while the runtime variant contains only the runtime to execute .NET applications.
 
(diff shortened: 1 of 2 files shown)


erikaheidi:
The CI is failing with the readme check, but when I run `monopod readme` here n othing happens - it actually changes a different readme (`images/doppler-kubernetes-operator/README.md`) so I am not sure what I am doing wrong here.

imjasonh (approved):

imjasonh:
```
Found a total of 1 issue(s). Please run the following at the root of this repo to fix: "monopod readme"
```

I'm not sure what the issue is but the tool should fix it for you 🤞

erikaheidi:
I updated monopod and was able to run the command, but it reverted the README to its original state. I think I am missing something @imjasonh 
cc @jamonation maybe I need to update some .tf file?

jamonation:
> The CI is failing with the readme check, but when I run `monopod readme` here n othing happens - it actually changes a different readme (`images/doppler-kubernetes-operator/README.md`) so I am not sure what I am doing wrong here.

Ah sorry about this @erikaheidi - it isn't obvious, but there are only two places to edit here, the `metadata.yaml` `compatibility_notes` section that can be multi-line markdown, and then in between the `<!--body:start-->...<!--body:end-->` section in `README.md`.

I know you're trying to edit the Download this image part, but that will always get overwritten. I think the least worst option would be to add the details about the two different images to the `compatibility_notes` section here.

jamonation:
@erikaheidi https://github.com/chainguard-images/images/pull/2325 is what you're after here I think?

erikaheidi:
> @erikaheidi #2325 is what you're after here I think?

Ok, that works for now, but we'll need to find a way to overwrite the download section, because there is no such image (cgr.dev/chainguard/dotnet). And this is the first info that shows up in the docs. This will probably happen with other images that have the same structure - they were broken down with different names than their parent folder.

erikaheidi:
Closing this since the proposed change won't pass CI, #2325 should be a temporary solution.

PR #2369, opened 2024-03-18
fix(cassandra-medusa): Add python-3.11 to extra packages

EyeCantCU:
Since cassandra-medusa enables no-depends, python-3.11 is not installed as an implicit runtime dependency so we need to explicitly require it

### DIFF
--- a/images/cassandra-medusa/config/main.tf
+++ b/images/cassandra-medusa/config/main.tf
@@ -9,6 +9,7 @@ variable "extra_packages" {
   default = [
     "py3-cassandra-medusa",
     "py3-cassandra-medusa-compat",
+    "python-3.11",
   ]
 }
 


images/cassandra-medusa/config/main.tf
```diff
@@ -9,6 +9,7 @@ variable "extra_packages" {
   default = [
     "py3-cassandra-medusa",
     "py3-cassandra-medusa-compat",
+    "python-3.11",
```
smoser, line 12: The change https://github.com/wolfi-dev/os/pull/15397 seems preferable to this one, and you should not need both.
Additionally, if you were going to specify a python package here, shouldn't it be the same as the one there (`python-3.11-base`)?

EyeCantCU: We need at least one default version of python in the container though and the package needs at least python-3.11-base to be functional after installing it otherwise we'd be providing a package with broken dependencies. If we explicitly ask for python-3.11 in the package, we break the ability to install other variants of python. No depends had to be enabled for cassandra-medusa due to an unrelated issue so we lost the dependency on so:libpython3~3.11. Installing python-3.11 currently installs python-3.11-base so we are still satisfying the dependency there while also ensuring the container has a default version of python available

If I'm overlooking something here, please let me know and I can get it fixed. Would definitely love to handle this in a better way

EyeCantCU:
We confirmed we just need the base 3.11 package for the virtual environment. Closing

PR #2386, opened 2024-03-20
Limit GITHUB_TOKEN permissions

jedsalazar:


mattmoor (approved):

jdolitsky (approved):

xnox:
@jdolitsky @mattmoor why did this pr switch from using `@main` to a pinned chainguard-dev/actions commit, which is not upgraded by any automation?

mattmoor:
I suspect @jedsalazar didn't realize that branches weren't supported by dependabot.  We can revert that aspect of this until we have a more serviceable solution.

PR #2411, opened 2024-03-27
[go] Only include openssh-client

wlynch:
I noticed while doing some image analysis that we're including `openssh-server` in the Go image. I don't see why this is needed, so following what the git image is doing and only include the client.


imjasonh (approved):

stormqueen1990 (approved):

mattmoor:
What's this waiting on?

PR #2436, opened 2024-04-02
minor update to Python Image README

SharpRake:
Per [this slack convo](https://chainguard-dev.slack.com/archives/C04PYHWPE1F/p1711726846991599) it was determined that we should document how to access the shell in the `python:latest-dev` image. This PR adds a line explaining how to do so and an example command one can run to open up the shell.

### DIFF
--- a/images/python/README.md
+++ b/images/python/README.md
@@ -38,6 +38,12 @@ docker pull cgr.dev/chainguard/python:latest
 docker pull cgr.dev/chainguard/python:latest-dev
 ```
 
+Note that in order to access the shell in the `python:latest-dev` image, you'll need to include an `--entrypoint` option, as in the following example.
+
+```sh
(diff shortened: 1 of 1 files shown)


images/python/README.md
```diff
@@ -38,6 +38,12 @@ docker pull cgr.dev/chainguard/python:latest
 docker pull cgr.dev/chainguard/python:latest-dev
 ```
 
+Note that in order to access the shell in the `python:latest-dev` image, you'll need to include an `--entrypoint` option, as in the following example.
+
+```sh
+docker run -it --entrypoint bin/bash chainguard/python:latest-dev
```
jdolitsky, line 44: I'd probably use absolute vs relative path in case workdir changes at any point:
```suggestion
docker run -it --entrypoint /bin/bash chainguard/python:latest-dev
```

PR #2445, opened 2024-04-04
Add iptables to istio-install-cni image

jamonation:
The istio-install-cni image needs `iptables` to function correctly: 

https://github.com/istio/istio/blob/master/cni/deployments/kubernetes/Dockerfile.install-cni#L16
https://github.com/istio/istio/blob/master/tools/istio-iptables/pkg/README.md

### DIFF
--- a/images/istio/configs/install-cni/main.tf
+++ b/images/istio/configs/install-cni/main.tf
@@ -7,6 +7,7 @@ terraform {
 variable "extra_packages" {
   description = "The additional packages to install (e.g. istio-install-cni, istio-cni...)"
   default = [
+    "iptables",
     "istio-install-cni",
     "istio-install-cni-compat",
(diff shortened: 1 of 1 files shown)


dlorenc (approved):
Is there a test we missed to catch this in the future?

jamonation:
I'm looking into that now @dlorenc - I'm not sure how istio uses iptables. I see it also emulates some ipsets functionality as well, and some external istio images include it, some don't

fraenkel:
Pre 1.21, istio-cni would delete pods that had CNI setup issues.
1.21, the default behavior switched to repair the CNI for a given pod.

https://github.com/istio/istio/blob/c70251eb23725a16f77e1f85730e380932a41d44/cni/pkg/repair/repaircontroller.go#L102

PR #2449, opened 2024-04-05
chore(imagetest/helm): add logs to Helm output

stormqueen1990:
Include a command to list logs from the pods deployed by a Helm chart to the `tflib/imagetest/helm` scripts.

### DIFF
--- a/.terraform.lock.hcl
+++ b/.terraform.lock.hcl
@@ -42,14 +42,14 @@ provider "registry.terraform.io/chainguard-dev/cosign" {
 }
 
 provider "registry.terraform.io/chainguard-dev/imagetest" {
-  version = "0.0.14"
+  version = "0.0.15"
   hashes = [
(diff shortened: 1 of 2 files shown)


joshrwolf:
I worry this is too much noise to always include. For some images, the pod logs are crazy. How about compromising with an opt in variable?

stormqueen1990:
> I worry this is too much noise to always include. For some images, the pod logs are crazy. How about compromising with an opt in variable?

I updated it, but the issue now is how would we turn it on when running on CI?

joshrwolf (approved):

found-it (dismissed):
One thing to note is this will shove logs into the `.diagnostic.detail` field. It doesn't look like there's a size restriction on the field but we'll certainly put that to the test

stormqueen1990:
> One thing to note is this will shove logs into the `.diagnostic.detail` field. It doesn't look like there's a size restriction on the field but we'll certainly put that to the test

@found-it it shouldn't always print in CI though, this is more for local debugging. Is there another way that this could become more suitable?

PR #2451, opened 2024-04-05
fix(conda): Initialize and enter base environment at startup

EyeCantCU:
Adds the base environment as well as a wrapper for initialization out of the box

Also runs image as root so that conda recognizes the base env exists

### DIFF
--- a/images/conda/config/template.apko.yaml
+++ b/images/conda/config/template.apko.yaml
@@ -2,18 +2,13 @@ contents:
   packages:
     - busybox
     - conda
+    - conda-base
+    - conda-wrapper
 
 accounts:
-  groups:
-    - groupname: nonroot
-      gid: 65532
-  users:
-    - username: nonroot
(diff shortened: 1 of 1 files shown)


images/conda/config/template.apko.yaml
```diff
@@ -2,6 +2,8 @@ contents:
   packages:
     - busybox
     - conda
+    - conda-base
+    - conda-wrapper
 
 accounts:
   groups:
```
pdeslaur, line 9: Do we still need to define the groups/users we run as root?

EyeCantCU: I'm not 100% certain on that. I followed the structure we currently use for other images that run as root. We could try without

EyeCantCU: Looks like we're good!

pdeslaur (approved):

PR #2462, opened 2024-04-09
migrate filebeat to imagetest

found-it:
filebeat sometimes fails its test with a conflicting CRD. imagetest runs in a sandbox which eliminates the conflicting CRD problem

### DIFF
--- a/images/filebeat/tests/main.tf
+++ b/images/filebeat/tests/main.tf
@@ -1,6 +1,7 @@
 terraform {
   required_providers {
-    oci = { source = "chainguard-dev/oci" }
+    oci       = { source = "chainguard-dev/oci" }
+    imagetest = { source = "chainguard-dev/imagetest" }
   }
 }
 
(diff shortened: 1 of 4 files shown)


images/filebeat/tests/main.tf
```diff
@@ -10,49 +11,60 @@ variable "digest" {
...
+
+  steps = [
+    {
+      name = "Helm install"
+      cmd  = module.helm.install_cmd
+    },
+    {
+      name  = "Test filebeat"
+      cmd   = <<EOF
+            export FB_NAMESPACE=default
```
stormqueen1990, line 60: Can we make this an environment variable in the harness perhaps?

images/filebeat/tests/test.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# TODO: Convert this to run with imagetest_harness_container when available
```
stormqueen1990, line 3: 🎉

images/filebeat/tests/main.tf
found-it: Good callout, honestly we don't actually have to set it anymore since the test just runs in the default namespace. I'll fix it up and remove this

stormqueen1990 (approved):

PR #2482, opened 2024-04-11
refactor istio to use imagetest_harness_k3s

jamonation:


jamonation:
This is stuck on creating the svc/istio-system-gateway via the `./test-gateway.sh` script. I'm not sure what the issue is.

jamonation:
Got the gateway test sorted, this is ready for a second set of 👀

images/istio/tests/operator/main.tf
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+module "helm" {
+  source = "../../../../tflib/imagetest/helm"
+
+  namespace = var.values.namespace
+
+  chart         = "istio-operator"
+  repo          = "https://stevehipwell.github.io/helm-charts/"
```
joshrwolf, line 28:
```suggestion
  repo          = "https://stevehipwell.github.io/helm-charts/"
```

can we use the official istio helm chart for this? or maybe even `istioctl`?

joshrwolf (approved):
I realize this is a port of the existing stuff, so no worries about the `stevehipwell` chart in this one.

thanks for doing this!

jamonation:
I looked around a bit for another operator chart, or a way to reuse the official ones but it doesn't seem like there are any with the right knobs and levers, thanks for checking!

PR #2483, opened 2024-04-12
powershell: add less package as dependency

kaniini:
Powershell uses `less -P` when paging, which is not supported by BusyBox less.


k4leung4 (approved):

kaniini:
Thanks @k4leung4!

PR #2558, opened 2024-04-24
Expand medusa test coverage, remove resource cleanup

mamccorm:
Related PR: https://github.com/wolfi-dev/os/pull/17569
Expand medua test coverage, and remove resource cleanup, which is not needed when using our new test harnes.

### DIFF
--- a/images/cassandra-medusa/tests/medusa-install.sh
+++ b/images/cassandra-medusa/tests/medusa-install.sh
@@ -2,42 +2,6 @@
 
 set -o errexit -o nounset -o errtrace -o pipefail -x
 
-# The CRDs which are left behind by the helm charts.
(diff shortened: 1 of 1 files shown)


images/cassandra-medusa/tests/medusa-install.sh
```diff
@@ -87,7 +51,7 @@ spec:
         size: 1
         storageConfig:
           cassandraDataVolumeClaimSpec:
-            storageClassName: standard
+            storageClassName: local-path
```
mamccorm, line 54: the test harness seems to use local-path as the default sc name, looks like this was missed in previous update

EyeCantCU (approved):

PR #2564, opened 2024-04-24
Updated NeMo overview, test scripts

smythp:
## Description

Updates the image overview and test scripts for NVIDIA NeMo. Adds a new script, tts.py, that will be used as the example script for now. Refactored and added missing nemo_quickstart.py script, but issues may need to be resolved before it can be used as the example script for this image.

## Notes

Added TTS example (tts.py) as current example script

Added missing nemo_quickstart.py code (requires more deps to run)

Added GPU testing section

Added test script command

Added resources

Changed description to a more technical description

This PR is related to docs and test scripts only.

References https://github.com/chainguard-dev/internal/issues/3393

References https://github.com/chainguard-dev/internal/issues/3749

### DIFF
--- a/tts.py
+++ b/tts.py
@@ -1,0 +1,42 @@
+from pathlib import Path
+import numpy as np
+from scipy.io.wavfile import write
+import torch
+from nemo.collections.tts.models.base import SpectrogramGenerator, Vocoder
+
+
+script_directory = Path(__file__).resolve().parents[0]
+
+if torch.cuda.is_available():
+    device = "cuda"
+else:
+    device = "cpu"
+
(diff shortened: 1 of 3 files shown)


srishtih (approved):

tts.py
```diff
@@ -0,0 +1,42 @@
+from pathlib import Path
```
jdolitsky, line 1: did we mean to check this in? @smythp

smythp: It should be in images/nemo/tests. Pushed a second commit to move it but a little too late. Will create a new PR to move it now.

PR #2570, opened 2024-04-25
Fix reaper tmpdir envrionment variables

found-it:
Looks like we do need the environment variables mentioned in https://github.com/chainguard-images/images/pull/2567

I fired up the container and checked the process and it's running with

```
    1 nonroot   0:56 java -Xms2147483648 -Xmx2147483648 -Djava.io.tmpdir= -cp /usr/local/lib/* io.cassandrareaper.ReaperApplication server /etc/cassandra-reaper/cassandra-reaper.yml
```

Which is setting the tmpdir to an empty string since the env is missing and the entrypoint tries to use it.

### DIFF
--- a/images/cassandra-reaper/config/template.apko.yaml
+++ b/images/cassandra-reaper/config/template.apko.yaml
@@ -22,6 +22,18 @@ entrypoint:
 cmd: cassandra-reaper
 
 paths:
+  - path: /var/tmp/cassandra-reaper
+    type: directory
+    uid: 65532
+    gid: 65532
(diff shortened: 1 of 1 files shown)


mritunjaysharma394 (approved):

mritunjaysharma394:
lgtm, verified it locally and upstream has that change too https://github.com/thelastpickle/cassandra-reaper/blob/1b4f8627f8e286261f4c4ab84f3aef3f87005e5f/src/server/src/main/docker/Dockerfile#L119

PR #2606, opened 2024-05-01
use init-upgrade

jdolitsky:


imjasonh (approved):

xnox:
@jdolitsky @imjasonh this is inverse of the intent.

intent was that infastrucutre, and local builds use the terraform lock file and do not randomly upgrade things.

otherwise.... there is no point in having terraform lock file committed at all. 

I.e. i was hoping init-upgrade is done in images digests CI updates only; such that upgrades are controlled.

Was this done because somebody submitted an upgrade which was done without `init-upgrade` and thus was missing linux-amd64 hashes again?

PR #2670, opened 2024-05-13
upload imagetest-logs directory

joshrwolf:
uses the new file based logging for `imagetest` to output and upload `imagetest` based logs.

this bypasses the strict logging required by terraform providers by dumping them to a file, which we'll postprocess with various downstream tooling

### DIFF
--- a/images/cert-manager/main.tf
+++ b/images/cert-manager/main.tf
@@ -39,6 +39,7 @@ module "config" {
   source   = "./config"
   name     = each.value.component
   suffix   = each.value.suffix
+
 }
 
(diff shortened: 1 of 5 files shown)


images/cert-manager/main.tf
```diff
@@ -39,6 +39,7 @@ module "config" {
   source   = "./config"
   name     = each.value.component
   suffix   = each.value.suffix
+
```
found-it, line 42:
```suggestion
```

found-it (approved):

PR #2676, opened 2024-05-14
make init-upgrade to get linux-amd64 hashes

xnox:
somehow i hoped this would be automated, but it is not.


smoser (commented):
one thing to do here would be to make a test that woudl fail in c-i if 'terraform init' made changes to this file.

xnox:
for reasons unknown to me, ci has been changed to do `make init-upgrade` meaning it ignores the lock file....

smoser (approved):

joshrwolf (approved):

PR #2677, opened 2024-05-14
migrate to imagetest

joshrwolf:
remove the `acceptance.sh` test and replace it with an inline that doesn't rely on a bindmount


images/vault/tests/main.tf
```diff
@@ -17,54 +18,69 @@ data "oci_string" "ref" {
...
+  description = "Basic functionality of the vault image."
+
+  steps = [
+    {
+      name = "Helm install vault"
+      cmd  = module.helm.install_cmd
+    },
+    {
+      name = "Acceptance test"
+      cmd  = <<EOF
```
smoser, line 68: This is some weird stuff. What processes this '<<EOF' ?
If it is passed to a shell, then you will get interpretation inside the body.
Ie, the 'kubectl exec' command will run _on the host_ and the output of that placed into the rendered document.

```bash
$ FOO=outside sh -c 'cat <<EOF
FOO=inside
echo $(echo $FOO)
EOF'

FOO=inside
echo outside
```

If you did it with double quotes around the EOF, then you get what you were after:

```bash
$ FOO=outside sh -c 'cat <<"EOF"
FOO=inside
echo $(echo $FOO)
EOF'

FOO=inside
echo $(echo $FOO)
```

smoser: or is '<<' a terraform template operator? link?

xnox (approved):

images/vault/tests/main.tf
```diff
@@ -17,54 +18,69 @@ data "oci_string" "ref" {
...
+apk add jq
+
+KEY=$(kubectl exec -n vault-system vault-0 -- vault operator init \
+  -key-shares=1 \
+  -key-threshold=1 \
+  -format=json | jq -r ".unseal_keys_b64[]")
```
smoser, line 74: Without explicitly stating 'bash' and 'set -o pipefail' this is just going to swallow errors.
I really just prefer doing this sort of thing in 2 steps.

```
out=$(kubectl exec -n vault-system vault-0 -- vault operator init \
  -key-shares=1 \
  -key-threshold=1 \
  -format=json) || { echo "kubectl exec vault failed: $out"; exit 1; }
 key=$(echo "$out" | jq -r ".unseal_keys_b64[]") || { echo "json parsing failed. doc=$out"; exit 1; }
  ```

File: images/vault/tests/main.tf [hunk 3]
```diff
@@ -17,54 +18,69 @@ data "oci_string" "ref" {
...
+
+kubectl exec -n vault-system vault-0 -- vault operator unseal $KEY
```
smoser, line 76: "KEY" should be quoted here.

smoser (changes requested):

images/vault/tests/main.tf
xnox: https://developer.hashicorp.com/terraform/language/expressions/strings#heredoc-strings

joshrwolf: `<<` is terraforms multiline string: https://developer.hashicorp.com/terraform/language/expressions/strings#heredoc-strings

this gets passed in as a raw string to `/bin/sh -c '...'`

images/vault/tests/main.tf
xnox: and what is the harm? invalid key? then unseal will fail, and test fails.

smoser: and the error message is hidden down below the thing that actually went wrong.

PR #2689, opened 2024-05-15
upgrade tf-apko, re-enable git alpine build

imjasonh:
The latest tf-apko build should be able to build images without hitting the git-lfs issue that was blocking Alpine-based `git` image builds before.

This undoes #2687 with the fix.


luhring (approved):

imjasonh:
The git image build passed, force merging past broken presubmit tests.

PR #2706, opened 2024-05-20
migrate apko test to imagetest (and more)

imjasonh:
Also:
- remove a deprecated `oci_string` from jitsucom test
- migrate the `TEMPLATE` repo to imagetest from helm and oci_exec_test


main.tf
```diff
@@ -80,6 +80,11 @@ provider "helm" {
 
 variable "newrelic_license_key" { default = "foo" } # set something valid to avoid targetted local runs
 
+module "blah" {
+  source            = "./images/blah"
+  target_repository = "${var.target_repository}/blah"
+}
+
```
found-it, line 87:
```suggestion
```

Is this supposed to be here?

images/TEMPLATE/tests/main.tf
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+locals { parsed = provider::oci::parse(var.digest) }
+
+module "helm" {
+  source = "../../../tflib/imagetest/helm"
+
+  name      = "TODO"
+  namespace = "TODO"
```
found-it, line 57: I usually leave resources in the default namespace for imagetest, do we want to recommend folks use `namespace` here or leave it off?

imjasonh: Oh that's a good idea. In the Old Way we had to put things in separate namespaces, but now it's a lot better. Done!

found-it (approved):

PR #2722, opened 2024-05-23
Expand tests, dynamically update CASSANDRA_VERSION var

mamccorm:
Expand tests, dynamically update CASSANDRA_VERSION var. Also went back and adjusted retries in cassandra test.

This image (mgmt-api-for-cassandra) cannot trigger cassandra tests, as there are other cassandra tests which are not compatible with this image, hence why copying.


images/management-api-for-apache-cassandra/main.tf
```diff
@@ -1,17 +1,35 @@
...
+data "apko_config" "for-getting-cassandra-version" {
+  config_contents = jsonencode({
+    contents = {
+      packages = ["cassandra"]
+    }
+  })
+}
+
+// Extract the cassandra version from the dummy config.
+locals {
```
mamccorm, line 18: This image requires 'CASSANDRA_VERSION' to be set as env var. Rather than hard code this, we grab it from the package. This isn't pretty, but we do similar in other places, so it's not new, and it means no more hard-coding CASSANDRA_VERSION

images/management-api-for-apache-cassandra/tests/run-tests.sh
```diff
@@ -0,0 +1,80 @@
...
+    docker logs "$CASSANDRA_CONTAINER"
+    docker stop "$CASSANDRA_CONTAINER"
+    docker rm "$CASSANDRA_CONTAINER"
+    exit 1
+  fi
+done
+
+# Clean up
+docker stop "$CASSANDRA_CONTAINER"
+docker rm "$CASSANDRA_CONTAINER"
```
joshrwolf, line 78:
```suggestion
docker rm -f "$CASSANDRA_CONTAINER"
```

fishing in the dark, I wonder if this is racing?

mamccorm: valid shout-out, actually we don't need these at all now the test is using the test harness right? so i'll remove both and not do any cleanup in the test and see how far that gets

mamccorm: didn't make a difference unfortunately but continuing to try a few more things

jdolitsky:

PR #2727, opened 2024-05-23
default check-sbom: true

imjasonh:
Spot-checking failed tests, they seem to be pre-existing test flakes/failures, and not related to this new check.

If there are any images with invalid SBOMs for any reason, we can unblock with `check-sbom = false` for that image, and address it later.


xnox (approved):
please notify engineering oncall when landing this.

PR #2756, opened 2024-05-30
boring-registry: get boring-registry from extra_packages

imjasonh:


xnox (approved):
And like redundant, as

```
variable "extra_packages" {
  description = "The additional packages to install (e.g. boring-registry)."
  default     = ["boring-registry"]
}
```

is already specified.

PR #2761, opened 2024-05-31
check-sbom: default to false again :(

imjasonh:
Undoes #2727 

Image builds seem to fail to install `spdx-tools-java`, and don't notify Slack as expected.


luhring (approved):

imjasonh:
This isn't going to pass all tests, and it passes enough that I believe it fixes the issue.

PR #2763, opened 2024-05-31
ensure jenkins works with plugins

joshrwolf:
I'm about the furthest thing from a jenkins knower, but apparently plugins are important.

This is ensure the scripts/dependencies needed for initializing, supporting, and downloading plugins are available in the image.

Most importantly, it ensures we're compatible with the upstream default helm chart installation. Previously, if you tried to install a vanilla jenkins installation we would bomb out for various plugin reasons.

### DIFF
--- a/images/jenkins/config/template.apko.yaml
+++ b/images/jenkins/config/template.apko.yaml
@@ -1,35 +1,44 @@
 contents:
   packages:
-    # jenkins and openjdk provided by extra_packages
+    # jenkins, jenkins-docker, and openjdk provided by extra_packages
 
 accounts:
   groups:
     - groupname: jenkins
-      gid: 65532
+      gid: 1000
   users:
     - username: jenkins
-      uid: 65532
-  run-as: 65532
-
-work-dir: /app
+      gid: 1000
+      uid: 1000
+      homedir: "/var/jenkins_home"
+  # Match upstream
+  run-as: 1000
 
 entrypoint:
-  command: jenkins.sh
+  command: "tini -- /usr/local/bin/jenkins.sh"
 
 environment:
   LANG: en_US.UTF-8
   JAVA_HOME: /usr/lib/jvm/java-17-openjdk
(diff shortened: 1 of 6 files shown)


joshrwolf:
leaving in draft until https://github.com/chainguard-dev/apko/pull/1146 is landed

found-it:
image

images/jenkins/config/template.apko.yaml
```diff
@@ -1,35 +1,45 @@
...
     - username: jenkins
-      uid: 65532
-  run-as: 65532
-
-work-dir: /app
+      gid: 1000
+      uid: 1000
+      homedir: "/var/jenkins_home"
+  # Match upstream
+  run-as: 1000
```
jamonation, line 16: Will this change be an issue for anyone using this with existing PVs that get remounted with new uid/gids?

joshrwolf: not if they're using the chart defaults, which were setting it to `1000` at runtime anyways

images/jenkins/config/template.apko.yaml
```diff
@@ -1,35 +1,45 @@
...
+      gid: 1000
   users:
     - username: jenkins
-      uid: 65532
-  run-as: 65532
-
-work-dir: /app
+      gid: 1000
+      uid: 1000
+      homedir: "/var/jenkins_home"
```
jamonation, line 14: CI and me both seem to be getting this message here? Should I be upgrading something?

```
│ yaml: unmarshal errors:
│   line 8: field homedir not found in type types.User
```

joshrwolf: working on rolling out all the necessary changes, this had to be added in apko https://github.com/chainguard-dev/apko/pull/1146

k4leung4 (approved):

jamonation (approved):
kubernetes credentials and everything, love it

PR #2773, opened 2024-05-31
cassandra-medusa: make grpc-health-probe configurable

pnasrat:
This makes grpc-health-probe an image build option


imjasonh (approved):

pnasrat:
@smoser I am working on an issue where these images haven't building/releasing currently so I am not sure if that is accurate.

PR #2789, opened 2024-06-03
tomcat imagetests

jamonation:


images/tomcat/tests/main.tf
```diff
@@ -1,32 +1,81 @@
 terraform {
   required_providers {
-    oci = { source = "chainguard-dev/oci" }
+    oci       = { source = "chainguard-dev/oci" }
+    imagetest = { source = "chainguard-dev/imagetest" }
   }
 }
 
+locals {
+  war_url = "https://tomcat.apache.org/tomcat-10.1-doc/appdev/sample/sample.war"
```
developer-guy, line 9: as it has version info in the URL (10.1) does that work with the upcoming versions of the Tomcat as well?

jamonation: It is a pretty simple hello world application, it works in 9.x, and here in 10. When 11 comes out, if it doesn't work then tests will break, but I expect it will be ok

developer-guy (approved):

PR #2810, opened 2024-06-05
fix(dex): revert extra_packages change

stormqueen1990:
Revert `extra_packages` change due to undesired side-effects.


imjasonh:
Blocking this for now while we move some things around in this repo.

PR #2816, opened 2024-06-06
GH actions cleanup

imjasonh:
- dependabot is only responsible for bumping actions commits
- most presubmit checks are gone, done in another repo -- all that's left is the check that a PR doesn't modify automated files
- withdrawal workflows are gone, done in another repo


jdolitsky (approved):

jdolitsky:
We will have to make `presubmit-roundup` no longer a required check

PR #2868, opened 2024-07-29
remove an environment variable from jellyfin that is no longer valid post 10.9.0

jsievenpiper:
An environment variable `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` supplied to the image can prevent syncing from occurring correctly.

https://github.com/jellyfin/jellyfin/issues/12009 provides a screenshot of documentation highlighting this, but I failed to find the original source. I can confirm removing it manually (by overriding the env var to empty) does seem to fix syncing 👍🏾


imjasonh:
Thanks for this contribution! We've adopted this change in the private mirror and it should be rolled out soonish.

PR #2869, opened 2024-07-29
Make go a base image (not application)

julienv3:
Fixes https://github.com/chainguard-dev/customer-issues/issues/1618


julienv3:
Ah, copied files, nvm!

PR #2871, opened 2024-07-30
docs: update grafana helm chart docs

jsievenpiper:
Updates the Grafana readme to properly specific where the image registry should go, otherwise the chart will attempt to pull from `docker.io/cgr.dev/chainguard/grafana`. Also added a little snippet if one would also like to leverage the k8s-sidecar image that you've built!

https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L99

Worth noting there is a `global` configuration for registry, but there are some images that do not have chainguard equivalents, so it isn't safe to use the global configuration if you use any of the images that chainguard doesn't have an equivalent for.


jdolitsky:
hello, and thank you for these documentation improvements. we have made the change internally and they should now be live at https://github.com/chainguard-images/images/blob/main/images/grafana/README.md

PR #2874, opened 2024-08-01
BugFix: Fixing a module call for var.target_repository in postgres helm compat image

liamd-rokt:
## Bug Fix

The Postgres Helm Compat terraform defintion had an issue in the call to test-latest.
This small fix should ensure that var.tagret_repository is passed to the module.


jdolitsky:
hello @liamd-rokt - thanks! Looks like we fixed this recently on our end. Closing for now

PR #2884, opened 2024-08-07
docs: fixes in squid-proxy docs

am97:
This PR makes the following changes:

- Fix testing section
  - A `curl` to localhost from another container wont work. The correct host would be 172.17.0.1 most of the time, but could be different in some setups. It's simpler to run the Curl command from the host.
- Fix container startup with custom config
  - By default, squid reads `/etc/squid.conf`, not `/etc/squid/squid.conf`
  - `http_port` must be provided, or startup will fail

Tested with current `cgr.dev/chainguard/squid-proxy:latest` (`cgr.dev/chainguard/squid-proxy@sha256:a5f31df17079ed15e40457285a59fa2119f08f73bc6675e70095c2cdbc76c3d2`)

### DIFF
--- a/images/squid-proxy/README.md
+++ b/images/squid-proxy/README.md
@@ -52,12 +52,12 @@ export PROXY_HOST="localhost"
 export PROXY_PORT="3128"
 export URL="http://example.com"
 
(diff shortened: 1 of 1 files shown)


jdolitsky:
hello, and thank you for these documentation improvements. we have made the change internally and they should now be live at https://github.com/chainguard-images/images/blob/main/images/squid-proxy/README.md

PR #2140, opened 2024-01-30
no world-writable dest path when running git container

Alestrix:
Add info on how to run the container with one's own UID so that making the destinaiton path world-writable becomes unnecessary.

Template does not apply, this is a documentaiton update.

PR #2912, opened 2024-10-10
no world-writable dest path when running git container

Alestrix:
Re-creation of #2140


jdolitsky:
Hi @Alestrix - thanks for the documentation change. We have made this internally and our automation has updated the files here: https://github.com/chainguard-images/images/commit/3ce35e9e4e42bbdec392ee8798ddf5866b0cf1ff#diff-1584119ed1e94171be1b98f5a57f1cbdfa1e31608738b7d14e1dada6b82623cd

Closing PR as there is now no diff

PR #2956, opened 2025-03-15
Ignore tj-actions/changed-files updates

egibs:
This Action has been compromised and we should avoid any updates.


mattlorimor (approved):

imjasonh (approved):

cpanato:
@egibs should we update to use another? we use this in other places

PR #2986, opened 2025-06-21
github: reclaim disk space for large builds

xnox:
Disk reclaim takes a long time. For images that run out of disk space
it is however needed.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -110,6 +110,51 @@ jobs:
         with:
           persist-credentials: false
 
+      # note this step takes a long time, thus only do it when we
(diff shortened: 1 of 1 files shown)


dannf (approved):
LGTM. Thank you for looking into this! Curious what algorithm you used to identify this set of paths. Might be good to leave that as a comment so future us's know how to update it.

xnox:
Oh, maybe doing rm like this inline is very fast compared to bootstrapping an imported action to run it.

https://github.com/chainguard-images/images/actions/runs/15797253936/job/44530888884

disk clean up took 20s.

It now has 56G available to build the image, let's see if it will work.

PR #3032, opened 2025-11-12
Add bit about setting LD_LIBRARY_PATH

Taffer:
Documenting the work-around for https://github.com/chainguard-dev/customer-issues/issues/2692


xnox (changes requested):
all changes should be made in images-private repository.

Either private or public image there, which will then be copied and published here.

PR #3079, opened 2026-04-13
ci(images): add parallel pipeline for image migration

uhryniuk:
### Notes for reviewer

We're moving towards using locked image configs to have reproducible builds for our public images.

Considering this, we are setting up a second release workflow which will build and publish the locked configs to a temporary register. Once the quality has been assured to the newly locked images, we will cutover the workflow and continue publishing to our public repository (`cgr.dev/chainguard`).

To be clear, we will have 2 parallel build pipelines for a temporary time until we can assert image quality and parity to the legacy TF-based counterparts.

### DIFF
--- a/.github/workflows/lock-release.yaml
+++ b/.github/workflows/lock-release.yaml
@@ -1,0 +1,236 @@
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'images/**/locked_config.json'
+  schedule:
+    - cron: "0 2 * * *"
+  workflow_dispatch:
+    inputs:
+      only:
+        description: "Specific image name to build"
+        type: string
+        required: false
+        default: ""
+
+concurrency:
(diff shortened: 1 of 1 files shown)


Rodrigo-Munoz-Salas (dismissed):

.github/workflows/lock-release.yaml
```diff
@@ -0,0 +1,190 @@
...
+        type: string
+        required: false
+        default: ""
+
+concurrency:
+  group: ${{ inputs.only || 'lock-release' }}
+  cancel-in-progress: false
+
+env:
+  TF_VAR_target_repository: cgr.dev/chainguard-cue-public
```
Rodrigo-Munoz-Salas: I think this needs to be `chainreg.biz/chainguard-cue-public`

Rodrigo-Munoz-Salas (approved):

PR #3088, opened 2026-04-22
docs(images): updating documentation files

Rodrigo-Munoz-Salas:
Refs: CON-1357

Updated documentation to the following (based on latest images changes):

1. README.md: Proper landing page. Explains what the repo is, describes the current build flow (locked_config.json → release.yaml → lmain.tf), adds a directory layout table, pull quickstart, and links to all supporting docs.
2. BUILDING.md (new): Customer-facing guide for reproducing a build locally using only public tooling (apko + jq). Covers the locked_config.json structure, a 3-command recipe to extract the config and build with apko, single-arch vs. multi-arch builds, and publishing to a personal registry.
3. BEST_PRACTICES.md: Removed stale references to the legacy build flow (HCL tflib/accts alternate examples throughout, the tflib/tagger legacy note, and the broken ./TERRAFORM.md link). All YAML/HCL dual-examples collapsed to YAML only (the format used in locked_config.json.)
4. WITHDRAWING_IMAGES.md: Fixed two broken workflow URLs that had a stray /os/ path segment.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,106 @@
+# Building Chainguard Images locally
+
+Every image in this repository ships a `locked_config.json` — a lock file that contains the fully resolved, pinned build inputs for each supported architecture. It embeds per-architecture [`apko`](https://github.com/chainguard-dev/apko) configurations (packages at exact versions, accounts, entrypoint, annotations, and so on) that our release pipeline feeds directly to the `apko` Terraform provider to produce the images at `cgr.dev/chainguard`.
+
+To build locally with the `apko` CLI, you first need to use `jq` to extract the raw apko config from the lock file — the CLI doesn't consume `locked_config.json` directly. The steps below walk through that.
+
+> `apko` and Chainguard's public package repository (`apk.cgr.dev/chainguard`) are enough to build any image in this repo — no Chainguard account or internal tooling required.
+
+## Prerequisites
+
+- [`apko`](https://github.com/chainguard-dev/apko?tab=readme-ov-file#installation)
+- [`jq`](https://jqlang.org/)
+- Docker (or any tool that can load an OCI tarball) — optional, only needed to run the image afterwards
+
+## What's inside `locked_config.json`
+
+The top-level shape is:
+
+```json
+{
+  "imageLocks": {
+    "<variant-name>": {
+      "configs": {
+        "amd64":  "{\"config\": { ...apko config for amd64... }}",
+        "arm64":  "{\"config\": { ...apko config for arm64... }}",
+        "index":  "{\"config\": { ...merged config for the index... }}"
+      },
+      "devConfigs": { ... },
+      "repo":  "public/<name>",
+      "tags":  ["latest", ...]
+    }
+  }
+}
+```
+
+Each entry under `imageLocks` is one publishable variant. The `configs` map holds a stringified JSON apko configuration per architecture — package list at pinned versions, accounts, entrypoint, annotations, paths, and so on. The release pipeline passes these directly to the `apko` Terraform provider; to use them with the `apko` CLI you extract the inner `.config` object with `jq` (see below).
+
+## Build an image in three commands
+
+The example below builds the `static` image. Swap `static` for any directory name under [`images/`](./images) to build a different image.
+
+```sh
+# 1. Extract the apko config for the desired variant.
+jq -r '.imageLocks["static-public"].configs.index | fromjson | .config' \
+  images/static/locked_config.json > static.apko.json
+
+# 2. Build with apko. --arch host limits to your current architecture;
+#    drop it to build the full multi-arch index.
+apko build static.apko.json cgr.dev/chainguard/static:local static.tar --arch host
+
+# 3. (Optional) Load into Docker and run.
+docker load < static.tar
+docker run --rm cgr.dev/chainguard/static:local
+```
+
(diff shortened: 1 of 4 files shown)


BUILDING.md
```diff
@@ -0,0 +1,90 @@
+# Building Chainguard Images locally
+
+Every image in this repository ships a `locked_config.json` file that fully describes how the image is assembled. This file is the input to [`apko`](https://github.com/chainguard-dev/apko), the same open-source tool our release pipeline uses. With `apko` and `jq` installed you can reproduce a build on your own machine.
```
uhryniuk: Not necessarily true, "This file is the input to apko...", when really it's a lock file that contains the inputs for apko to build that image for different archiectures.

Mentioning that it's a lock file is important, since we can use it to generate the exact images that are found in `cgr.dev/chainguard`, with the exception of images that require private APK repos at build time.

Rodrigo-Munoz-Salas: Yeah I meant that, I thought it sounded clear when saying "`locked_config.json` file that fully describes how the image is assembled". Technically this file is mainly for users to understand how to use apko to play around locally

uhryniuk: I'd push back on this. Technically the file makes no assumptions about users. It has a unique schema, that requires explanation to fully be used with `apko`. I'd agree `locked_config.json` in name implies clearly that it is a lock file but it doesn't distinguish for _what_. Is it for `apko`? Is it for a TF module? When it can actually be used for both.

It'd be more clear to state the lock file is directly used with the `apko` TF provider, but to access the config to build directly with the `apko` CLI tool, we need to use `jq` to pull out the raw YAML configuration.

TLDR; I'd consider being more explicit and evident here.

BUILDING.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Publishing to your own registry
+
+`apko publish` pushes directly to a registry instead of writing a tarball:
+
+```sh
+apko publish static.apko.json ghcr.io/<you>/static:local
+```
+
+You'll need to be authenticated to the destination registry (`docker login`, `gh auth`, etc.). This produces an image identical in content to the one at `cgr.dev/chainguard/static` for the same lock file, minus the SBOM and cosign attestations that our release pipeline adds on top.
```
uhryniuk: It wouldn't hurt to give an example with `crane` to verify the digests.

WITHDRAWING_IMAGES.md
```diff
@@ -23,6 +23,6 @@ Sometimes an image repo needs to be removed entirely because it was erroneously
 To do so:
 
 - Add the repo basename that need to be removed to `withdrawn-repos.txt`
-- Run the ["Withdraw Repos"](https://github.com/chainguard-images/images/os/blob/main/.github/workflows/withdraw-repos.yaml) workflow on GitHub.
+- Run the ["Withdraw Repos"](https://github.com/chainguard-images/images/blob/main/.github/workflows/withdraw-repos.yaml) workflow on GitHub.
```
uhryniuk, line 26: Any clue why this changed?

README.md
```diff
@@ -1,13 +1,57 @@
...
+
+```
+images/<name>/
+├── locked_config.json   # Pinned apko config per architecture (+ tags, repo)
+├── README.md            # User-facing documentation for the image
+├── tests/               # End-to-end tests
+└── metadata.yaml        # Image metadata
+```
+
+`locked_config.json` is the source of truth for what goes into the published image. It contains a fully resolved apko configuration — exact package versions, accounts, entrypoint, environment, and annotations — for each supported architecture. Because the package versions are pinned, builds are reproducible: the same `locked_config.json` produces the same image content.
```
uhryniuk, line 28: May be worth noting here or somewhere else that we won't actually get the same digests because of the time created label.

BUILDING.md
uhryniuk: Nvm, digests won't match from since the time created will be different. Perhaps there is another option for us.

WITHDRAWING_IMAGES.md
Rodrigo-Munoz-Salas: It changed a while back, documentation didn't get updated ([history](https://github.com/chainguard-images/images/commits/main/.github/workflows/withdraw-repos.yaml))

BUILDING.md
Rodrigo-Munoz-Salas: I can just add the command to run probably to compare digests towards the live repo --> e.g. `diff <(crane digest cgr.dev/chainguard/<image>:<tag>) <(crane digest <local-repo>/<image>:<tag>)`

BUILDING.md
Rodrigo-Munoz-Salas: Got it!! Updating it now

README.md
```diff
@@ -1,13 +1,57 @@
...
+## Building an image locally
+
+You can reproduce a build on your own machine using `apko` and the `locked_config.json` checked into this repo. See [`BUILDING.md`](./BUILDING.md) for a step-by-step example.
+
+## Repository layout
+
+| Path | Purpose |
+| --- | --- |
+| `images/<name>/` | Per-image build config, tests, and README |
+| `lock-release/` | Terraform driver that consumes `locked_config.json` and publishes |
```
uhryniuk: lock-release is now gone

Rodrigo-Munoz-Salas: updated it!

uhryniuk (approved):

PR #3091, opened 2026-04-23
ci: refactor release to use root main.tf

uhryniuk:
### Notes for reviewers

Now that we are deprecating and removing the legacy TF configured image build system, we can final promote our lock-release build sytem to the root of our project.

Moving forward we are able to build the image solely with the main.tf and the locked_config.json files.


.github/workflows/release.yaml
Rodrigo-Munoz-Salas, line 43: There are lint issues here

uhryniuk: Ack. onit bruv

Rodrigo-Munoz-Salas:
there is a few more lint issues reported by action lint in lines 151, 212, and 222

Rodrigo-Munoz-Salas (approved):

PR #3115, opened 2026-07-02
fix(ci): GitHub Actions security hardening

stevebeattie:
This PR applies GitHub Actions hardening fixes across the workflows in
`.github/workflows/`, surfaced by static analysis (zizmor).

## What this changes

- **`persist-credentials: false` on checkout steps.** The affected jobs don't
  push back to the repo, so the auto-persisted `GITHUB_TOKEN` in `.git/config`
  is unnecessary — dropping it narrows the blast radius of later steps.
- **Template-injection fix.** `workflow_dispatch` input expressions that were
  interpolated directly into `run:` blocks are moved into `env:` and referenced
  as quoted shell variables, so the shell — not the workflow templater — handles
  them.
- **Repo-level zizmor config.** Disables cosmetic pedantic-only rules and sets a
  dependabot cooldown threshold, so future scans stay focused on substantive
  findings.

## Test plan

- `zizmor .github/` — clean after these changes.
- `actionlint` — workflows still parse with no errors.

Refs: PSEC-923

### DIFF
--- a/.github/workflows/digestabot.yaml
+++ b/.github/workflows/digestabot.yaml
@@ -23,6 +23,8 @@ jobs:
         egress-policy: audit
 
     - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      with:
+        persist-credentials: false
 
     - uses: octo-sts/action@f603d3be9d8dd9871a265776e625a27b00effe05 # v1.1.1
       id: octo-sts
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -42,6 +42,8 @@ jobs:
       - id: shard
         name: Shard
         shell: bash
+        env:
+          INPUTS_ONLY: ${{ inputs.only }}
         run: |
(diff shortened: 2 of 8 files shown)


egibs (approved):

.github/workflows/digestabot.yaml
```diff
@@ -23,6 +23,8 @@ jobs:
         egress-policy: audit
 
     - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      with:
+        persist-credentials: false
```
cpanato, line 27: i think here needs to be true as we need to commit later

.github/workflows/release.yaml
```diff
@@ -36,10 +36,14 @@ jobs:
           egress-policy: audit
 
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
```
cpanato, line 40: need to test this one, maybe need to be true as well

PR #3129, opened 2026-09-10
Harden Withdraw Repos workflow: safety check, push trigger, no swallowed failures (CON-2596)

tomgeorge:
Hardens the dispatch-only **Withdraw Repos** workflow (zero runs since 2023) to the standard of stereo's live private-catalog port. Fix item 5 of [CON-2596](https://linear.app/chainguard/issue/CON-2596) / [internal-dev#26825](https://github.com/chainguard-dev/internal-dev/issues/26825).

## Changes

- Deletion loop moved to `hack/withdraw-repos.sh`, ported from stereo's `containers/hack/withdraw-repos.sh`. Each entry resolves to a repo UIDP under the public `chainguard` org before deletion: bare names match direct children exactly (no interactive-selector hang in CI), missing repos skip with a warning, and slashed entries (nested repos) error and fail the run — nested repos are an invalid state, cleaned up manually by UIDP.
- `|| true` removed. Failures accumulate and the script exits non-zero — one bad repo no longer aborts the rest, and a red run now means something.
- `repocountbysource` safety check: refuses to withdraw a repo that customer repos sync from. `skip_safety_check` dispatch input bypasses it (same knob as stereo).
- Push trigger on `withdrawn-repos.txt` (`main` only): the mirror sync from stereo runs the withdrawal live. Manual dispatch still defaults to `dry_run: true`.
- Slack alert on failure via the existing `DISTROLESS_SLACK_WEBHOOK`.
- `if: github.repository == 'chainguard-images/images'` fork guard; `WITHDRAWING_IMAGES.md` now says to edit the file in stereo, not here.

Kept `egress-policy: audit` (repo convention); tightening egress on a never-exercised workflow belongs in its own PR.

Validated: `bash -n`; workflow YAML parsed; script smoke-tested with stubbed `chainctl`/`curl` — dry-run default, failure accumulation + exit 1, not-found skip, safety gate blocks and fails, bypass works, slashed entry rejected with non-zero exit.

## Deferred

The end state is one owner: extend the `stereo-withdraw-containers` identity to the public org and retire this legacy pair. Tracked in [CON-2865](https://linear.app/chainguard/issue/CON-2865). The operator runbook for the CON-2596 cleanup lives on chainguard-dev/stereo#322388.

### DIFF
--- a/.github/workflows/withdraw-repos.yaml
+++ b/.github/workflows/withdraw-repos.yaml
@@ -1,15 +1,27 @@
+name: Withdraw Repos
+
 on:
   workflow_dispatch:
     inputs:
       dry_run:
         type: boolean
         default: true
-        description: If true, just log
+        description: Execute as dry run, only log details
+      skip_safety_check:
+        type: boolean
+        default: false
+        description: Bypass the customer-repo source count safety check
+  push:
+    branches: [main]
+    paths:
+      - "withdrawn-repos.txt"
 
 permissions: {}
 
 jobs:
   withdraw:
+    if: github.repository == 'chainguard-images/images'
+
     runs-on: ubuntu-latest
 
     permissions:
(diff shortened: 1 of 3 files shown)


tomgeorge:
Reopening: withdrawals stay in this repo, so in-place hardening is the plan of record again (optional). Note: this PR enforces bare-names-only, so merge it only after the imagetest UIDP cleanup runs through the legacy workflow. *(via Claude Code)*

tomgeorge:
Closing for now: not needed for the CON-2596 cleanup — the remaining imagetest deletion (chainguard-dev/stereo#330627) runs through the workflow as-is, and this PR's bare-names-only enforcement would reject those UIDP entries. Long-term ownership of public-org withdrawals is tracked in CON-2865; reopen if in-place hardening becomes the plan of record.

